PR #95, opened 2023-03-06
User story/1642 read input files

randy-seng:
OP#1684
OP#1642


OTAnalytics/application/datastore.py
```diff
@@ -12,14 +14,56 @@ def parse(self, file: Path) -> list[Track]:
         pass
 
 
-@dataclass(frozen=True)
-class Video:
-    path: Path
+class VideoReader(ABC):
```
randy-seng, line 17: Not sure if this should stay here 🤔

OTAnalytics/plugin_video_processing/video_reader.py
```diff
@@ -0,0 +1,36 @@
+from pathlib import Path
+
+import numpy as np
+from moviepy.video.io.VideoFileClip import VideoFileClip
+
+from OTAnalytics.application.datastore import VideoReader
+
+
+class FrameDoesNotExistError(Exception):
```
randy-seng, line 9: Should we extract it to another module? If so which layer should it belong to?

OTAnalytics/application/datastore.py
briemla: I will check this with the code I wrote on the other PR.

OTAnalytics/domain/track.py
```diff
@@ -1,33 +1,100 @@
...
+        frame (int): the frame that the detection belongs to.
+        occurrence (datetime): the time of the detection's occurence.
+        input_file_path (Path): absolute path to otdet that the detection belongs to
+        at the time of its creation.
+        interpolated_detection (bool): whether this detection is interpolated.
+        track_id (int): the track id this detection belongs to.
+    """
+
+    classification: str = Field(alias=ottrk_format.CLASS)
+    confidence: float = Field(gt=0, le=1, description="must be between range [0,1]")
```
briemla, line 44: Range `(0,1]` which would be: `gt=0` and `le=1` or range `[0,1]` which would be `ge=0` and `le=1`?

briemla: As description I suggest: "must be in range"

OTAnalytics/domain/track.py
```diff
@@ -1,33 +1,100 @@
...
+    x: float = Field(ge=0, description="must be greater equal than 0")
+    y: float = Field(ge=0, description="must be greater equal than 0")
+    w: float = Field(ge=0, description="must be greater equal than 0")
+    h: float = Field(ge=0, description="must be greater equal than 0")
+    frame: int = Field(gt=0, description="must be greater than 0")
     occurrence: datetime
     input_file_path: Path
-    interpolated_detection: bool
-    track_id: int
+    interpolated_detection: bool = Field(alias=ottrk_format.INTERPOLATED_DETECTION)
```
briemla, line 52: Why do some fields use aliases and some not?

OTAnalytics/domain/track.py
```diff
@@ -1,33 +1,100 @@
...
+    track_id: int = Field(
+        gt=0,
+        description="Track ID must be greater equal than 0",
```
briemla, line 55: greater or equal is `ge`?

OTAnalytics/domain/track.py
```diff
@@ -1,33 +1,100 @@
...
+    """
+
+    id: int = Field(gt=0, description="id must be a number greater than 0")
     detections: list[Detection]
 
+    @validator("detections")
+    def detections_must_be_in_right_order(
+        cls, detections: list[Detection]
+    ) -> list[Detection]:
+        if detections != sorted(detections, key=lambda det: det.frame):
```
briemla, line 83: We can sort by occurrence instead of frame number. Might be more robust in the long run, when detections of multiple videos can be in the same track.

OTAnalytics/plugin_parser/otvision_parser.py
```diff
@@ -1,9 +1,89 @@
+import bz2
+import json
```
briemla, line 2: Should we use `ujson` instead of `json` as in OTVision for better performance?

OTAnalytics/plugin_video_processing/video_reader.py
briemla: I would place it here for the moment. It might move into the application code (`datastore.py`).

OTAnalytics/plugin_video_processing/video_reader.py
```diff
@@ -0,0 +1,36 @@
+from pathlib import Path
+
+import numpy as np
```
briemla, line 3: Do we need aliases for imports? We can just use numpy.ndarray on usage.

tests/domain/test_track.py
```diff
@@ -0,0 +1,102 @@
...
+
+import pytest
+from pydantic import ValidationError
+
+import OTAnalytics.plugin_parser.ottrk_dataformat as ottrk_format
+from OTAnalytics.domain.track import Detection, Track
+
+
+@pytest.fixture
+def valid_det_dict() -> dict:
```
briemla, line 12: det is detection?
If yes, can we use detection instead?

pyproject.toml
```diff
@@ -1,5 +1,8 @@
+[project]
+name = "OTAnalytics"
+
 [build-system]
-requires = ["setuptools"]
+requires = ["setuptools", "hatchling"]
```
briemla, line 5: Why do we need hatchling?

briemla (changes requested):

OTAnalytics/domain/track.py
randy-seng: Thanks, that's correct! I will fix that!

OTAnalytics/domain/track.py
randy-seng: Right the description doesn't match. My intention was have Track ID  start from 1 as reflected in the ottrk file.

Do we want track id to be in `[0, +inf)` or `[1, +inf)`?

pyproject.toml
randy-seng: Actually we don't need it. I just picked it up in some basic pyrpoject.toml example without questioning it. My bad 😅 I'll remove it

OTAnalytics/domain/track.py
randy-seng: That's a good question! 

First up why I decided to use aliases at all. `INTERPOLATED_DETECTION` in the ottrk file is `interpolated-detection`. In Python, hyphens are not allowed to be used in variable names. Thus, it is spelled using snake_case. I thought it would be neat to use pydantic's serialisation features at some point in the future. At this moment I only use it in our tests. Thus, defining aliases allows us to easily de- or serialise a detection without the need to care about making the decision to use snake_case or hyphen variant of as a key in this use case. 

Now to answer your question, I wasn't thinking ahead. A change in the data format of ottrk might require changes here as well. So, defining aliases for the other attributes might be an appropriate thing to do.

OTAnalytics/domain/track.py
randy-seng: I set the constraint to be `track_id  in [1,0)`. I will change it if requested~

requirements-dev.txt
```diff
@@ -1,9 +1,9 @@
-black
-flake8
+black==23.1.0
```
randy-seng, line 1: I pinned those dependencies also used by pre-commit so that they are always the same

OTAnalytics/domain/track.py
briemla: We should discuss if we want third party library dependencies in domain and application code. If not, we can move the hyphen vs. underscore in names into a plugin. Our domain code should not know anything about serialisation because it is not part of our domain.

briemla (commented):
Looks good. I just made a few changes.

briemla (approved):
Looks good. I just made a few changes.

PR #97, opened 2023-03-08
task/1682-create-gui-to-define-sections

martinbaerwolff:
OP#1682


briemla:
The connection between datastore and UI is currently a draft and will be updated in op#829.

randy-seng (approved):
Looks good. Just pushed a quick fix.

PR #99, opened 2023-03-13
Feature/1683 define eventlist format

randy-seng:
OP#1683

### DIFF
--- a/OTAnalytics/domain/event.py
+++ b/OTAnalytics/domain/event.py
@@ -135,11 +135,6 @@ class SectionEventBuilder(EventBuilder):
         self.direction_vector: Optional[DirectionVector2D] = None
         self.event_type: Optional[EventType] = None
 
-    def reset(self) -> None:
-        self.section_id = None
-        self.direction_vector = None
-        self.event_type = None
-
     def add_section_id(self, section_id: str) -> None:
         self.section_id = section_id
 
@@ -175,5 +170,4 @@ class SectionEventBuilder(EventBuilder):
             direction_vector=self.direction_vector,
             video_name=detection.input_file_path.name,
         )
-        self.reset()
         return section_event
--- a/OTAnalytics/domain/event.py
+++ b/OTAnalytics/domain/event.py
@@ -1,22 +1,26 @@
+import re
+from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from datetime import datetime
 from enum import Enum
+from pathlib import Path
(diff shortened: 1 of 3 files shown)


OTAnalytics/domain/event.py
```diff
@@ -1,2 +1,156 @@
...
+
+    Raises:
+        ValueError: vehicle_id < 1
+        ValueError: frame_number < 1
+
+    Args:
+        road_user_id (int): the road user id involved with this event. It must be
+        greater equal one.
+        road_user_type (RoadUserType): the road user type involved with this event.
+        hostname (str): the OTCamera hostname that the video is associated with.
```
randy-seng, line 88: How do we treat videos not recorded by an OTCamera?

OTAnalytics/domain/event.py
```diff
@@ -1,2 +1,177 @@
...
+        raise InproperFormattedFilename(f"Could not parse {file_path.name}.")
+
+
+class SectionEventBuilder(EventBuilder):
+    def __init__(self) -> None:
+        self.section_id: Optional[str] = None
+        self.direction_vector: Optional[DirectionVector2D] = None
+        self.event_type: Optional[EventType] = None
+
+    def reset(self) -> None:
```
briemla, line 136: Why do we need a reset method?
Can we not just create a new Builder?

OTAnalytics/domain/geometry.py
```diff
@@ -0,0 +1,78 @@
...
+    def _validate(self) -> None:
+        if len(self.coordinates) < 3:
+            raise ValueError(
+                (
+                    "Number of coordinates to define a valid polygon must be "
+                    f"greater equal three, but is {len(self.coordinates)}"
+                )
+            )
+
+        if self.coordinates[0] != self.coordinates[-1]:
```
briemla, line 40: This means, we need at least 3 coordinates. If exaclty 3 are given, coordinate 1 and must be same. That would be a line. So the number of coordinates must either be greater or equal 4 or coordinates[0] will automatically added at the end to close the polygon.

randy-seng: That is correct! Thank you! I'll fix that~

OTAnalytics/domain/event.py
randy-seng: I did it just in case some one tries to reuse the builder. But of course we can just use a new one

briemla (approved):

PR #110, opened 2023-03-24
Feature/1774 create events when a road user enters a scene for the first and the last time

randy-seng:
OP#1774


briemla (approved):
Looks good.

PR #114, opened 2023-03-27
Feature/1873 assign a max class to a track depending on the classes of the detections

randy-seng:
OP#1873


briemla (approved):
Looks good so far.

PR #125, opened 2023-04-05
[#1925] Add section data to serialized event list

randy-seng:
https://openproject.platomo.de/work_packages/1925

OP#1925


briemla (approved):
Looks good.

PR #129, opened 2023-04-09
Feature/1914 create UI to define line sections

martinbaerwolff:
OP#1914


randy-seng (approved):
Looking good! I decoupled the GUI from the application and fixed a bug where the section list is not updated when removing a section.

briemla (approved):
Looks good. A huge step forward for our ui progress.

PR #136, opened 2023-04-12
[#1872] Calculate direction vector based on previous and next detection

randy-seng:
https://openproject.platomo.de/work_packages/1872

OP#1872


briemla (approved):
Looks good so far.

PR #152, opened 2023-04-20
User story/2002 as a user, i want to choose which offset is used to plot the tracks

briemla:
OP#2002


randy-seng (dismissed):
Looks good! I did some small refactorings and decoupled the `TracksFrame` class from the `OTAnalyticsApplication`

randy-seng (approved):

PR #153, opened 2023-04-20
User story/2003 as a user, i want to use the offset of the selected sections to plot tracks

briemla:
OP#2003


randy-seng (approved):
Looking good!

PR #172, opened 2023-04-28
User story/1956 define a visualization plugin to plot track information

briemla:
OP#1956


randy-seng (approved):
Looks good! Just pushed fixes to typo in docstrings and wrong type annotation.

PR #175, opened 2023-04-29
User story/1951 define an extensible filter element with implementations for filtering Track and DataFrame

randy-seng:
OP#1951 Define an extensible filter element
OP#1945 Filter tracks or dataframe by classes
OP#1947 Filter tracks or dataframe by date range

### DIFF
--- a/OTAnalytics/application/track_filter.py
+++ b/OTAnalytics/application/track_filter.py
@@ -1,0 +1,127 @@
+from datetime import datetime
+from typing import Callable, Iterable, Optional
+
+from OTAnalytics.domain.filter import Filter, FilterBuilder, FilterBuildError, Predicate
+from OTAnalytics.domain.track import Track
+
+
+class TrackPredicate(Predicate[Track, bool]):
+    """A predicate that takes in a `Track` to test against.
+
+    Args:
+        predicate (Callable[[Track], bool] | None): Optional to supply a complex
+            predicate. If none is supplied, the predicate implemented in this class
+            will be used.
+    """
+
+    def __init__(self, predicate: Callable[[Track], bool] | None = None) -> None:
+        super().__init__(predicate)
+
+    def conjunct_with(self, other: Predicate[Track, bool]) -> Predicate[Track, bool]:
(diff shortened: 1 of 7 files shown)


OTAnalytics/application/track_filter.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def apply(self, data: Iterable[Track]) -> Iterable[Track]:
+        return [datum for datum in data if self._predicate.test(datum)]
+
+
+class TrackFilterBuilder(FilterBuilder):
+    """A builder used to build a `TrackFilter`."""
+
+    def __init__(self) -> None:
+        self._complex_predicate: Optional[Predicate[Track, bool]] = None
```
briemla, line 103: We might turn this into a list of predicates and conjunct them in the `build` method.

briemla: This would reduce the complexity of the `add_...` methods. Another option is to extract the `if-else`-Block of the `add_...` methods into an internal `conjunct` method which is called by both `add_...` methods.

At the moment I would favor the internal `conjunct` method.

OTAnalytics/plugin_filter/dataframe_filter.py
```diff
@@ -0,0 +1,162 @@
...
+        return [datum[self._predicate.test(datum)] for datum in data]
+
+
+class DataFrameIsWithinDate(DataFramePredicate):
+    """Checks if the DataFrame rows are within a date range.
+
+    Args:
+        column_name (str): the DataFrame column name to apply the predicate to
+        start_date (datetime): the start date of the date range
+        end_date (datetime): the end date of the date range
```
briemla, line 72: See comment below about inclusive and exclusive dates. The documentation should state the final decission.

briemla (commented):
Looks quite good. Found one inconsistency which should be fixed. I fixed a bunch of typos and small issues. Please update your branch before further changes.

briemla (changes requested):
Looks quite good. Found one inconsistency which should be fixed. I fixed a bunch of typos and small issues. Please update your branch before further changes.

briemla (approved):

PR #176, opened 2023-05-01
Let line sections be defined by a list of multiple coordinates

randy-seng:
OP#2064


OTAnalytics/plugin_parser/otvision_parser.py
```diff
@@ -366,17 +366,14 @@ def _parse_line_section(self, data: dict) -> Section:
...
                 section.RELATIVE_OFFSET_COORDINATES,
-                section.START,
-                section.END,
             ],
         )
         section_id = self._parse_section_id(data)
         relative_offset_coordinates = self._parse_relative_offset_coordinates(data)
-        start = self._parse_coordinate(data[section.START])
-        end = self._parse_coordinate(data[section.END])
+        coordinates = self._parse_coordinates(data)
```
briemla, line 373: Do we need a format conversion?
There exist section files with `start` and `end` coordinates. Should we convert them?
This might be a topic for todays daily.

briemla (changes requested):
Looks good so far. The section files in tests/data and in data need to be converted. At the moment, only one otflow file is converted.

briemla (approved):

briemla (approved):

PR #178, opened 2023-05-03
User-story/2062-create-gui-element-to-define-multilinesections

martinbaerwolff:
OP#2062


OTAnalytics/plugin_ui/customtkinter_gui/line_section.py
```diff
@@ -202,14 +205,22 @@ def update(self, coordinate: tuple[int, int], event_type: str) -> None:
...
+        if event_type == "left_mousebutton_up":
+            print("left_mousebutton_up")
             self.geometry_builder.add_coordinate(coordinate)
+        elif (
+            self.geometry_builder.number_of_coordinates() >= 1
+            and event_type == "mouse_motion"
+        ):
+            self.geometry_builder.add_temporary_coordinate(coordinate)
+        elif self.geometry_builder.number_of_coordinates() >= 2 and (
+            event_type in {"right_mousebutton_up", "return"}
```
briemla, line 217: Why a dictionary and not a list: `[]` or tuple: `()` ?

briemla (commented):
Looks good so far. I could not reproduce the bug shown last time. I have only one question.

OTAnalytics/plugin_ui/customtkinter_gui/line_section.py
martinbaerwolff: Given the definition of a set, it seems to be behave a dict without values :)

briemla: That's right, a set is in python like a dict without values. I have to lower by Java skills for this one ;-)

briemla (approved):

PR #182, opened 2023-05-12
User story/1946 create UI to filter tracks by date

randy-seng:
OP#1946


OTAnalytics/plugin_ui/customtkinter_gui/frame_filter.py
```diff
@@ -0,0 +1,503 @@
...
+
+        if not self._has_valid_time_format(name, value):
+            self._display_invalid_validation_info(self._build_error_message_for(name))
+            widget.configure(border_color=COLOR_RED)
+            return True
+
+        widget.configure(border_color=COLOR_GREEN)
+        self._clear_validation_info()
+        return True
+
```
briemla, line 454: The if structure in this method is semantically the same as in the `_valida_date_format` method: Validate a date or time object and show the result as color.

However, the logic in `_validate_date_format` is positive: `if self._viewmodel.validate_date(value):` which is easy to read. The logic in `_validate_time_format` is negative: `if not ...` which is harder to read.

Can you change the structure of the methods to be more or less the same?

tests/OTAnalytics/application/test_state.py
```diff
@@ -145,3 +153,63 @@ def test_update_image(self) -> None:
...
+        tracks_metadata.notify_tracks([])
+        assert tracks_metadata.first_detection_occurrence == first_detection.occurrence
+        assert tracks_metadata.last_detection_occurrence == third_detection.occurrence
+
+        mock_get_all_track_detections.assert_called_once()
+
+    def test_get_all_track_detections(
+        self, first_detection: Mock, second_detection: Mock
+    ) -> None:
+        track = Mock(spec=Track).return_value
```
briemla, line 206: Should the `return_value` part be added to `track.detections` instead?

briemla (commented):
Looks good so far.

briemla (changes requested):
Looks good so far. Only a few comments to check.

OTAnalytics/domain/filter.py
```diff
@@ -149,10 +167,50 @@ def build_filter(self, filter_builder: FilterBuilder) -> Filter:
         if self.classifications:
             filter_builder.add_has_classifications_predicate(self.classifications)
 
-        if self.start_date and self.end_date:
-            filter_builder.add_starts_at_or_after_date_predicate(self.start_date)
+        if self.date_range.start_date and self.date_range.end_date:
```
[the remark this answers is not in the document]
randy-seng: You are correct we should not limit `start_date`

OTAnalytics/domain/filter.py
```diff
@@ -149,10 +167,50 @@ def build_filter(self, filter_builder: FilterBuilder) -> Filter:
...
+        self._by_date_filter_setting: Optional[DateRange] = None
+        self._by_classification_filter_setting: Optional[list[str]] = None
+
+    def save_by_date_filter_setting(self, filter_element: FilterElement) -> None:
+        self._by_date_filter_setting = filter_element.date_range
+
+    def save_by_classification_filter_setting(
+        self, filter_element: FilterElement
+    ) -> None:
+        self._by_classification_filter_setting = filter_element.classifications.copy()
```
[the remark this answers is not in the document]
randy-seng: Actually I was just taking precautions for unwanted mutations of the list.
The reason why I didn't do the same for the date filter was because of `date_range` being an immutable/frozen dataclass. But I think it's better to return a copy to avoid confusion

OTAnalytics/application/state.py
```diff
@@ -338,3 +342,64 @@ def notify_sections(self, sections: list[SectionId]) -> None:
...
+
+    Args:
+        TrackListObserver (TracListObserver): extends the TrackListObserver interface
+        track_repository (TrackRepository): the track repository
+    """
+
+    def __init__(self, track_repository: TrackRepository) -> None:
+        self._track_repository = track_repository
+        self._first_detection_occurrence: Optional[datetime] = None
+        self._last_detection_occurrence: Optional[datetime] = None
```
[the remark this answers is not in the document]
randy-seng: It would definitely not hurt us to have them as OptionalObervableProperties~

OTAnalytics/plugin_ui/customtkinter_gui/frame_filter.py
randy-seng: Will do!

tests/OTAnalytics/application/test_state.py
randy-seng: When trying to modify the mock objects members as you suggested will raise an error.
It's kind of hard to explain. Mocking methods will work as expected. But when it comes to mocking the objects members it is a little bit different. In this case the `Mock(spec=Track).return_value` is needed and  will return a mock instance that you can edit.

See: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.return_value

briemla (approved):

briemla (approved):

PR #187, opened 2023-05-15
Bug/2179-file-ending-for-otflow-file-is-not-automatically-appended-on-windows

martinbaerwolff:
OP#2179


briemla (approved):
Default file extension was missing on save. I added it.

PR #190, opened 2023-05-16
Feature/2113 implement basic UI

briemla:
OP#2113
OP#2118


randy-seng (approved):
Looking good! I updated some docstrings and extended a test case.

PR #191, opened 2023-05-17
Feature/2122 add flow as domain object

briemla:
OP#2122


OTAnalytics/application/datastore.py
```diff
@@ -4,6 +4,7 @@
 from typing import Iterable, Optional, Tuple
 
 from OTAnalytics.domain.event import Event, EventRepository
+from OTAnalytics.domain.flow import Flow, FlowId, FlowRepository
```
martinbaerwolff, line 7: Why do we not need observer classes for flows as we do for sections?

OTAnalytics/domain/flow.py
```diff
@@ -0,0 +1,43 @@
...
+        start: Section,
+        end: Section,
+        distance: float,
+    ) -> None:
+        self.id: FlowId = id
+        self.start: Section = start
+        self.end: Section = end
+        self._distance: float = distance
+
+    def distance(self) -> float:
```
martinbaerwolff, line 28: How can we overwrite the distance? Simply rebuild the flow for now?

OTAnalytics/domain/flow.py
```diff
@@ -0,0 +1,43 @@
...
+        return self._distance
+
+
+class FlowRepository:
```
martinbaerwolff, line 32: How can we delete a flow? Simply rebuild the whole repository for now?

OTAnalytics/domain/flow.py
```diff
@@ -0,0 +1,43 @@
...
+        return self.id
+
+
+class Flow:
+    def __init__(
+        self,
+        id: FlowId,
+        start: Section,
+        end: Section,
+        distance: float,
```
martinbaerwolff, line 21: Should we prevent negative values?

tests/OTAnalytics/domain/test_flow.py
```diff
@@ -0,0 +1,40 @@
...
+
+import pytest
+
+from OTAnalytics.domain.flow import Flow, FlowId, FlowRepository
+from OTAnalytics.domain.section import Section
+
+
+@pytest.fixture
+def flow() -> Flow:
+    distance = 1.0
```
martinbaerwolff, line 11: If we decide to prevent negative values, add a test for that

OTAnalytics/domain/flow.py
```diff
@@ -0,0 +1,43 @@
...
+    def serialize(self) -> str:
+        return self.id
+
+
+class Flow:
+    def __init__(
+        self,
+        id: FlowId,
+        start: Section,
+        end: Section,
```
martinbaerwolff, line 20: Should we prevent passing something other than a Section?

We could also think about prevent passing the same section as both start and end (but I also can think of a use case of u-turns, so maybe this is not necessary)

tests/OTAnalytics/domain/test_flow.py
```diff
@@ -0,0 +1,40 @@
...
+    start = Mock(spec=Section)
```
martinbaerwolff, line 12: If we decide to prevent passing Non-Sections, also test for that?

OTAnalytics/application/datastore.py
briemla: To update the ui on added, removed or renamed flows.

OTAnalytics/domain/flow.py
briemla: At this point yes, generate a new flow and save it to the repository. After OP#2123 is merged, the distance is editable.

martinbaerwolff (approved):

PR #206, opened 2023-06-01
dont try to build a line between two detections that are at the same point

arminkollascheck:
OP#2330

…point (otherwise ValueError: Coordinates define a closed line)

Only added one if-statement. Doesnt seem like this needed additional changes at other parts.


randy-seng:
@arminkollascheck This problem occurs because our domain does not  allow a `Line` geometry to be closed.
Thus, the error is being raised by trying to convert a closed `Track` to a `Line`.
I have removed this constraint and added some test cases. Would you please check if this problem has been resolved now?

briemla (approved):
Looks good.

PR #210, opened 2023-06-06
Bug/2262 when editing a section, a copy of the section with the same name is created

briemla:
OP#2262

### DIFF
--- a/OTAnalytics/plugin_ui/customtkinter_gui/dummy_viewmodel.py
+++ b/OTAnalytics/plugin_ui/customtkinter_gui/dummy_viewmodel.py
@@ -27,6 +27,7 @@ from OTAnalytics.domain.section import (
     NAME,
     RELATIVE_OFFSET_COORDINATES,
     LineSection,
+    MissingSection,
     Section,
     SectionId,
     SectionListObserver,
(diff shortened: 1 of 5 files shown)


OTAnalytics/plugin_ui/customtkinter_gui/dummy_viewmodel.py
```diff
@@ -297,6 +293,32 @@ def set_new_section(self, data: dict, coordinates: list[tuple[int, int]]) -> Non
         print(f"New line_section created: {line_section.id}")
         self._update_selected_section(line_section.id)
 
+    def __validate_section_information(
+        self, data: dict, coordinates: list[tuple[int, int]]
```
randy-seng, line 297: Maybe use a more specific name for parameter `data`? I.e. `meta_data`.
If so, maybe do the same for function `update_section_coordinates`?

randy-seng (commented):
Looks good! Just one question~

OTAnalytics/plugin_ui/customtkinter_gui/dummy_viewmodel.py
briemla: Hmm, the data contains everything except the coordinates. We could summarize it as metadata :-)

randy-seng (approved):

PR #211, opened 2023-06-07
User-story/2272 content of side panel or main panel is not show fully

briemla:
OP#2272

### DIFF
--- a/OTAnalytics/plugin_ui/customtkinter_gui/frame_filter.py
+++ b/OTAnalytics/plugin_ui/customtkinter_gui/frame_filter.py
@@ -407,7 +407,7 @@ class DateRow(CTkFrame):
 
     def _get_widgets(self) -> None:
         self.validation_info_label = CTkLabel(
-            master=self, text="", width=400, anchor=ANCHOR_WEST
+            master=self, text="", width=200, anchor=ANCHOR_WEST
         )
         self.name_label = CTkLabel(
             master=self, text=self._name, anchor=ANCHOR_WEST, width=40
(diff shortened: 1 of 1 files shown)


OTAnalytics/plugin_ui/customtkinter_gui/frame_filter.py
```diff
@@ -461,18 +461,29 @@ def _get_widgets(self) -> None:
...
 
     def _place_widgets(self) -> None:
-        self.name_label.grid(row=0, column=0, padx=PADX, pady=PADY, sticky=STICKY_WEST)
-        self.date_entry.grid(row=0, column=1, padx=PADX, pady=PADY, stick=STICKY)
-        self.hour_entry.grid(row=0, column=2, padx=PADX, pady=PADY, stick=STICKY)
-        ColonLabel(master=self).grid(row=0, column=3)
-        self.minute_entry.grid(row=0, column=4, padx=PADX, pady=PADY, stick=STICKY)
-        ColonLabel(master=self).grid(row=0, column=5)
-        self.second_entry.grid(row=0, column=6, padx=PADX, pady=PADY, stick=STICKY)
         if self._place_validation_below:
```
martinbaerwolff, line 464: Why two versions?

martinbaerwolff (approved):

OTAnalytics/plugin_ui/customtkinter_gui/frame_filter.py
briemla: In OTAnalytics, we need all ui elements in a row, because we show two times the same widget (e.g. in filter as from and to dates). In OTConfig, we want to have the widgets aligned with lower horizontal space requirements. Therefore, the columns and rows of the widgets differs.

PR #220, opened 2023-06-13
User story/2292 load preview image of video into memory

briemla:
OP#2292

### DIFF
--- a/tests/OTAnalytics/plugin_parser/test_otvision_parser.py
+++ b/tests/OTAnalytics/plugin_parser/test_otvision_parser.py
@@ -685,6 +685,19 @@ class TestCachedVideoParser:
         if isinstance(parsed_videos[1], CachedVideo):
             assert parsed_videos[1].other == video2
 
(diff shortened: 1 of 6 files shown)


tests/OTAnalytics/plugin_parser/test_otvision_parser.py
```diff
@@ -627,6 +630,78 @@ def test_serialize_events(
         assert event_list_file.exists()
 
 
+class TestCachedVideo:
+    def test_cache_frames(self, test_data_tmp_dir: Path) -> None:
```
randy-seng, line 634: Can we also add tests that the `CachedVideo` classes' `get_path` and `to_dict` methods are properly delegated to `other` as well?

randy-seng (commented):
Looks good!! Just one comment~

tests/OTAnalytics/plugin_parser/test_otvision_parser.py
briemla: Fixed :-)

randy-seng (approved):

PR #221, opened 2023-06-13
User story/2297 make distance of flows optional

briemla:
OP#2297


OTAnalytics/plugin_ui/customtkinter_gui/dummy_viewmodel.py
```diff
@@ -540,7 +540,7 @@ def add_flow(self) -> None:
             name = flow_data[FLOW_NAME]
             new_from_section_id = SectionId(flow_data[START_SECTION])
             new_to_section_id = SectionId(flow_data[END_SECTION])
-            distance = float(flow_data[DISTANCE])
+            distance = flow_data.get(DISTANCE, None)
```
randy-seng, line 543: It is possible to pass distance of 0 when creating a flow and thus a exception is being raised when creating a `Flow` instance in l.544.

randy-seng (commented):
Found a bug, please check~

randy-seng (approved):

PR #229, opened 2023-06-15
Feature/2192-add-cancel-button-to-metadata-popup-for-flows

martinbaerwolff:
OP#2192


briemla (approved):
Looks nice. Good work with the order of the buttons on mac and other OSs

PR #230, opened 2023-06-15
User story/2138 allow sectionstate to select a list of sections

randy-seng:
OP#2138
OP#2139


OTAnalytics/plugin_ui/customtkinter_gui/frame_flows.py
```diff
@@ -69,6 +66,21 @@ def _place_widgets(self) -> None:
     def action_buttons(self) -> list[CTkButton]:
         return self._action_buttons
 
+    def set_selection(self, flow_ids: list[str]) -> None:
+        pass
```
briemla, line 70: In `frame_sections.py` a `NotImplementedError` is raised. Should it be the same behaviour? Or should we remove the method completely?

briemla (commented):
Looks good. Only a small question.

briemla (commented):
Looks good. Only a small question.

OTAnalytics/plugin_ui/customtkinter_gui/frame_flows.py
randy-seng: Yes, it is not being used so we can remove method completely.

briemla (approved):

PR #243, opened 2023-06-21
Feature/2358 Show message box before closing the application

briemla:
OP#2358

### DIFF
--- a/OTAnalytics/plugin_ui/customtkinter_gui/gui.py
+++ b/OTAnalytics/plugin_ui/customtkinter_gui/gui.py
@@ -24,6 +24,18 @@ class ModifiedCTk(CTk):
         **kwargs: Any,
     ) -> None:
         super().__init__(**kwargs)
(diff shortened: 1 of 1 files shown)


martinbaerwolff (approved):
Looks good! Did you already add "close on Escape" to the Infobox?

Otherwise I could do this. With #248 I also added a FrameOkCancel to reuse in all Toplevel windows. This also has the switched-buttons-on-Mac feature. I would also add this to Infobox. Should we wait for this #248 to be merged and add it here or shoild

briemla:
> Looks good!
> 
> Did you already add "close on Escape" to the Infobox? Otherwise we could do this here.
> 
> With #248 I also added a `FrameOkCancel` to reuse in all Toplevel windows. This also has the `switched-buttons-on-Mac` feature. I would also add this to Infobox. Should we wait for #248 to be merged and add it with this PR or should I open another issue?

We should open another issue for this.

PR #248, opened 2023-06-22
Feature/2127-add-ui-element-to-export-counts

martinbaerwolff:
OP#2127


OTAnalytics/plugin_ui/customtkinter_gui/toplevel_template.py
```diff
@@ -1,45 +1,27 @@
...
-        )
-        if ON_MAC:
-            ok_column = 1
-            cancel_column = 0
-        else:
-            ok_column = 0
-            cancel_column = 1
-        self.button_ok.grid(row=0, column=ok_column, padx=PADX, pady=PADY)
-        self.button_cancel.grid(row=0, column=cancel_column, padx=PADX, pady=PADY)
-
```
briemla, line 42: Why is the Ok/Cancel frame removed from the template?
This results in duplicated code, because every implementation of `ToplevelTemplate` has to instantiate the frame_ok_cancel. It might be easier to have a "content" frame in `ToplevelTemplate` which is filled with elements from derived classes.

OTAnalytics/plugin_ui/customtkinter_gui/toplevel_template.py
```diff
@@ -1,45 +1,27 @@
...
+        title: str,
         initial_position: tuple[int, int],
         **kwargs: Any,
     ) -> None:
         super().__init__(**kwargs)
-        self._get_frame_ok_cancel()
+        self.title(title)
         self._set_ok_cancel_bindings()
         self._set_initial_position(initial_position)
-        self._set_focus()
```
briemla, line 20: Why do we not centrally set the focus to the toplevel?

briemla (changes requested):
Looks good so far. Only a few duplication questions.

OTAnalytics/plugin_ui/customtkinter_gui/toplevel_template.py
martinbaerwolff: Done :)

OTAnalytics/plugin_ui/customtkinter_gui/toplevel_template.py
martinbaerwolff: Done :)

briemla (approved):
Looks good. I changed only a few things and fixed a bug.

PR #249, opened 2023-06-22
User story/2087 run intersection for selected section

randy-seng:
OP#2087


OTAnalytics/domain/section.py
```diff
@@ -41,28 +42,16 @@ def notify_sections(self, sections: list[SectionId]) -> None:
...
-        """
-        Listen to events.
-
-        Args:
-            observer (SectionChangedObserver): listener to add
-        """
-        self.observers.add(observer)
+        super().__init__()
 
     def notify(self, value: SectionId) -> None:
```
briemla, line 56: Can we also move the notify method into the super class with a second generic type?

This would replace a lot of `...Subject` classes with just one.

OTAnalytics/domain/observer.py
```diff
@@ -0,0 +1,23 @@
+from abc import ABC
+from typing import Generic, TypeVar
+
+OBSERVER = TypeVar("OBSERVER")
+
+
+class Registrable(ABC, Generic[OBSERVER]):
```
briemla, line 7: We might add a second generic to reduce duplication of `...Subject` classes further :-) See comment in `section.py`.

OTAnalytics/domain/track.py
```diff
@@ -91,13 +92,13 @@ def notify(self, track_id: Optional[TrackId]) -> None:
...
+class TrackListSubject(Registrable[TrackListObserver]):
     """
     Helper class to handle and notify observers
     """
 
     def __init__(self) -> None:
-        self.observers: set[TrackListObserver] = set()
+        super().__init__()
 
     def register(self, observer: TrackListObserver) -> None:
```
briemla, line 103: This method can be removed. It only delegates to the super method.

OTAnalytics/domain/track.py
```diff
@@ -106,7 +107,7 @@ def register(self, observer: TrackListObserver) -> None:
         Args:
             observer (TrackListObserver): listener to add
         """
-        self.observers.add(observer)
+        super().register(observer)
 
     def notify(self, tracks: list[TrackId]) -> None:
```
briemla, line 112: See comment in `observer.py` and `section.py` for a replacement of this method.

OTAnalytics/plugin_ui/customtkinter_gui/dummy_viewmodel.py
```diff
@@ -437,6 +448,7 @@ def load_tracks(self) -> None:
         print(f"Tracks files to load: {track_files}")
         track_paths = [Path(file) for file in track_files]
         self._application.add_tracks_of_files(track_files=track_paths)
+        self._application.intersect_tracks_with_sections()
```
briemla, line 451: Is it necessary to run the intersection also in the next method (`load_sections`)?

briemla (changes requested):
Looks nice. Only a few questions.

briemla (approved):

PR #258, opened 2023-06-27
Feature/1906-create-gui-elements-to-define-area-sections

martinbaerwolff:
OP#1906


OTAnalytics/plugin_ui/customtkinter_gui/dummy_viewmodel.py
```diff
@@ -769,13 +815,31 @@ def _draw_arrow_for_selected_flows(self) -> None:
...
 
+    def _get_section_refpt_calculator(
+        self, section: Section
+    ) -> SectionRefPointCalculator:
+        if isinstance(section, LineSection):
+            return InnerSegmentsCenterCalculator()
+        elif isinstance(section, Area):
+            return GeometricCenterCalculator()
+        else:
+            raise ValueError("section has to be a LineSection or an Area, but isnt")
```
briemla, line 841: See comment in line 674

briemla (changes requested):

OTAnalytics/plugin_ui/customtkinter_gui/dummy_viewmodel.py
briemla: We could also do something like
```python
calculator_map: dict[Type, SectionRefPointCalculator] = {
      LineSection, InnerSegmentsCenterCalculator(),
      Area, GemetricCenterCalculator(),
      }
        dummy.get(type(section))
```

The dictionary could be setup in the `__init__` method. This would allow to add future types by adding a new entry into the dictionary instead of changing the code. (See Open Closed Principle for more about that: https://de.wikipedia.org/wiki/Open-Closed-Prinzip )

briemla (approved):

PR #259, opened 2023-06-27
feature/2194-visualization-should-not-filter-tracks-shorter-than-30-frames

randy-seng:
OP#2194

### DIFF
--- a/OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
+++ b/OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
@@ -87,7 +87,7 @@ CLASS_ORDER = [
 ]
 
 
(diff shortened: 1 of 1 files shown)


OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
```diff
@@ -248,15 +248,15 @@ def _filter_tracks(
         Returns:
             DataFrame: filtered by classes, time and number of images
         """
-        track_df[
+        filtered_df = track_df[
```
randy-seng, line 251: Because of the missing assignment of the `track_df` since the operation is not inplace, the tracks were not filtered to begin with. I fixed that here.

briemla (approved):

PR #261, opened 2023-06-27
Feature/2068-as-a-user,-i-want-to-export-the-event-list-as-csv-and-excel

martinbaerwolff:
OP#2068


OTAnalytics/plugin_prototypes/eventlist_exporter/eventlist_exporter.py
```diff
@@ -0,0 +1,168 @@
...
+    def build(self) -> pd.DataFrame:
+        self._convert_occurence_to_seconds_since_epoch()
+        self._split_columns_with_lists()
+        self._add_section_names()
+        return self._df
+
+    def _convert_occurence_to_seconds_since_epoch(self) -> None:
+        epoch = pd.Timestamp("1970-01-01")
+        occurence = pd.to_datetime(self._df[OCCURRENCE])
+        self._df[f"{OCCURRENCE}_sec"] = (occurence - epoch).dt.total_seconds()
```
briemla, line 39: The occurrence of an event should already have a `...timestamp...` method to convert it to a timestamp. We should use the same conversion at all places.

OTAnalytics/plugin_prototypes/eventlist_exporter/eventlist_exporter.py
```diff
@@ -0,0 +1,168 @@
...
+        default_factory=lambda: [
+            EventListExcelExporter(),
+            EventListCSVExporter(),
+            # EventListPrinter(),
+            # EventListDataFramePrinter(),
+        ]
+    )
+
+    def get_exporter_by_name(self, name: str) -> EventListExporter:
+        for exporter in self.AVAILABLE_EXPORTERS:
```
briemla, line 165: Store the available exporters in a dictionary and use `...get(name)` to get one.

briemla (changes requested):

OTAnalytics/plugin_prototypes/eventlist_exporter/eventlist_exporter.py
martinbaerwolff: This would mean unpacking the (large) eventlist dataframe, calling the timestamp function on the occurence and creating the dataframe again. Or using an apply method on the dataframe. This would be much slower than using the builtin pd.Timestamp(). This is a prototype - until the core of OTAnalytics is faster, I would like to leave it as it is, but add a comment here. Do you agree?

martinbaerwolff: Or do we already create a separate dataframe for the eventlist internally?

briemla: Not now, it will be done after profiling.

briemla: > This would mean iterating over the (large) eventlist. Or using an apply method on the dataframe. This would be much slower than using the builtin pd.Timestamp(). This is a prototype - until the core of OTAnalytics is faster, I would like to leave it as it is, but add a comment here. Do you agree?

True, I mixed it with the detections where the timestamp is a property of the objects and is also available in the dataframe as column.

briemla (dismissed):

briemla (approved):

PR #265, opened 2023-07-04
Feature/advanced track data frame cache

KueblerJelle:
Refactored CachedPandasTrackProvider:
Now cache is not invalidated upon every call of notify_tracks (e.g. if a few additional tracks are added).
Every change to the track repository is noticed (notified) and directly mirrored in the cache:

- added tracks are converted to DataFrame and appended to the cache

- existing tracks in the cache that have an updated version of it are also updated

Also added new tests for the various possible track changes.


OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
```diff
@@ -283,18 +295,85 @@ def _convert_tracks(self, tracks: Iterable[Track]) -> DataFrame:
...
+        Returns:
+            DataFrame: the dataframe representation of the given tracks.
+        """
+        return super()._convert_tracks(tracks)
+
     def notify_tracks(self, tracks: list[TrackId]) -> None:
         """Take notice of some change in the track repository.
-        Resets the cached dataframe, so it will be recomputed.
+
+        Remove cached tracks matching any given id.
```
randy-seng, line 317: Tracks with `TrackId`s that already exist in the cache should not be simply discarded. 

A `Track` with an unique `TrackId` can be continuous and thus appear in more than one ottrk file.
In that case the ottrk parser will take the existing `Track` and merge it with the new `Track` with the same id and update the track repository.

Thus an already existing TrackId being passed to the `notify_tracks` method indicates that there was actually an update of the respective track. In this case the correct behaviour would be to remove the track in the cache and replace it with the updated one.

In the future there are plans to have separate observers for removing and adding to a repository.

tests/OTAnalytics/plugin_prototypes/track_visualization/test_track_viz.py
```diff
@@ -103,12 +121,82 @@ def test_get_data_reset_cache(
...
+
+        provider.notify_tracks([self.track_2.id, self.track_1.id])
+        self.check_expected_ids(provider, [self.track_2, self.track_1])
+
+    def test_notify_update_delete(self) -> None:
+        self.create_tracks()
+        provider = self.set_up_provider([self.track_1, self.track_2], [])
+
+        provider.notify_tracks([self.track_1.id])
+        self.check_expected_ids(provider, [self.track_2])
```
randy-seng, line 185: See comment in `track_viz.py` l.317~

tests/OTAnalytics/plugin_prototypes/track_visualization/test_track_viz.py
```diff
@@ -81,20 +82,37 @@ def test_get_data_empty_track_repository(self) -> None:
 
 
 class TestCachedPandasTrackProvider:
-    def test_get_data_reset_cache(
-        self,
-    ) -> None:
-        id = TrackId(1)
+    def create_tracks(self) -> None:
```
randy-seng, line 85: We could replace this method with pytest fixtures for each track created here.
Pytest will provide them via dependency injection when declaring them as parameters in the test method.

```python
    @pytest.fixture
    def track_1(self) -> Track:
        return self.set_up_track(1)

    @pytest.fixture
    def track_2(self) -> Track:
        return self.set_up_track(2)
    
    def test_foo(self, track_1: Track) -> None:
        assert 1 == track_1.id.id

**Review by randy-seng — CHANGES_REQUESTED** [review]
Looking good! Just a couple of question~

File: OTAnalytics/plugin_prototypes/track_visualization/track_viz.py [hunk 4]
```diff
@@ -331,18 +343,77 @@ def _convert_tracks(self, tracks: Iterable[Track]) -> DataFrame:
...
+            track
+            for t_id in track_ids
+            if (track := self._datastore._track_repository.get_for(t_id))
+        ]
+
+    def _cache_without_existing_tracks(self, track_ids: list[TrackId]) -> DataFrame:
+        """Filter cached tracks.
+
+        Only keep those not matching the ids in the given list of track_ids.
+        Return `None` if the current cache is empty.
```
briemla, line 403: Do we really return `None` if all are filtered or is the DataFrame just empty?

OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
```diff
@@ -331,18 +343,77 @@ def _convert_tracks(self, tracks: Iterable[Track]) -> DataFrame:
...
+            track_ids (list[TrackId]): ids of tracks to be removed from cache.
+
+        Returns:
+            DataFrame : filtered cache.
+        """
+        if self._cache_df.empty:
+            return self._cache_df
+
+        track_id_nums = [t.id for t in track_ids]
+        df = self._cache_df
```
briemla, line 415: Why do we need a temporal variable?

OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
```diff
@@ -331,18 +343,77 @@ def _convert_tracks(self, tracks: Iterable[Track]) -> DataFrame:
...
+        return df[~df[track.TRACK_ID].isin(track_id_nums)]
```
briemla, line 416: Does this return a filtered view on the DataFrame or create a new DataFrame without the filtered elements?

If it is only a filtered view, the DataFrame will only increase over time. `df.drop[df[track.TRACK_ID].isin(track_id_nums)]` might reduce the memory footprint.

OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
randy-seng: You are correct. A DataFrame is always returned.

briemla (approved):

PR #266, opened 2023-07-05
Feature/2081 show progress for all actions

randy-seng:
OP#2081
OP#1933

### DIFF
--- a/OTAnalytics/application/progress.py
+++ b/OTAnalytics/application/progress.py
@@ -17,7 +17,33 @@ class SimpleCounter(Counter):
         return self._value
 
 
-class NotifyableProgressbar(Progressbar):
+class AutoIncrementingProgressbar(Progressbar):
+    def __init__(
+        self,
+        sequence: Sequence,
+        counter: Counter,
+        notify: Optional[Callable[[], None]] = None,
+    ) -> None:
+        self._sequence = sequence
+        self._counter = counter
+        self._notify = notify
+        self._iterator = iter(self._sequence)
+        self._counter.reset()
+
+    def __iter__(self) -> Iterator:
+        self._counter.reset()
+        self._iterator = iter(self._sequence)
+        while True:
+            try:
+                yield next(self._iterator)
+                self._counter.increment(1)
+                if self._notify:
+                    self._notify()
+            except StopIteration:
+                return
+
+
+class ManualIncrementingProgressbar(Progressbar):
     def __init__(
         self,
         sequence: Sequence,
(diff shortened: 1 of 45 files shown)


OTAnalytics/application/progress.py
```diff
@@ -0,0 +1,78 @@
...
+        self._counter.reset()
+        self._iterator = iter(self._sequence)
+        return self
+
+    def __next__(self) -> Any:
+        next_element = next(self._iterator)
+        if self._notify:
+            self._notify()
+        return next_element
+
```
briemla, line 69: Why do we need a `__next__` method here but not in the automatic version?

OTAnalytics/plugin_ui/customtkinter_gui/toplevel_progress.py
```diff
@@ -1,34 +1,51 @@
...
         super().__init__(**kwargs)
         self.title(title)
-        self._initial_message = initial_message
-        self._initial_progress = initial_progress
+        self._description = description
         self._get_widgets()
         self._place_widgets()
         self._set_focus()
         self.protocol("WM_DELETE_WINDOW", self._on_cancel)
         self.overrideredirect(True)  # TODO: Test on Windows
```
briemla, line 46: The MinimalInfoBox uses a direct call to tkinter instead of the wrapper. Is this also an issue here?

OTAnalytics/plugin_ui/customtkinter_gui/toplevel_progress.py
```diff
@@ -40,25 +57,112 @@ def _set_focus(self) -> None:
...
+    def __call__(
+        self, sequence: Sequence, description: str, unit: str
+    ) -> AutoIncrementingProgressbar:
+        counter = SimpleCounter()
+        self._popup_builder.add_counter(counter)
+        self._popup_builder.add_description(description)
+        self._popup_builder.add_unit(unit)
+        self._popup_builder.add_total(len(sequence))
+        popup = self._popup_builder.build()
+        progressbar = AutoIncrementingProgressbar(
```
briemla, line 165: Directly return the progressbar?

OTAnalytics/application/progress.py
randy-seng: The automatic version is basically a `Generator` in disguise. And every `Generator`  is an `Iterator` Whereas the manual version implements the `Iterator` interface.
I could have implemented the manual version as a generator as well. The reason why I didn't was due to the fact that `Generator`s are not pickable which is needed for `multiprocessing`.

OTAnalytics/plugin_ui/customtkinter_gui/toplevel_progress.py
randy-seng: There had been no complications so far with this widget, but definitely better to call it directly as well.

briemla (approved):

PR #274, opened 2023-07-20
bug/2387 After editing a flow´s properties, it remains highlighted on the canvas but is deselected in the list

martinbaerwolff:
OP#2387


briemla (changes requested):
See the updated description in OP

briemla (approved):

PR #277, opened 2023-07-21
Bug/2533 generate flows button is not disabled on start

briemla:
OP#2533


martinbaerwolff (approved):
Looks good. During testing, I found two other, quite similar bugs (OP#2535 and OP#2536)

PR #286, opened 2023-07-28
User-story/2342-clean-up-padding-in-inner-frames

martinbaerwolff:
OP#2342

PR #287, opened 2023-07-28
OP#2571: reorganize frame navigation in tab view

KueblerJelle:
Addressing bug OP#2571:

Reorganize frame navigation to make all widgets visible and clickable on small screens.

Transform FrameNavigation from CTKFrame to CTkTabView.
Add project frame, input tabview, config tabview and analysis frame as separate tabs.
Use project tab as initial tab.

Resulting GUI:
tab_solution


martinbaerwolff (commented):
We fixed the bug that not all widgets were visible on small screens in #286

martinbaerwolff:
We fixed the bug that not all widgets were visible on small screens in #286

PR #291, opened 2023-08-02
Bug/2540 save eventlist create file with empty file name but file extension

briemla:
OP#2540


martinbaerwolff (approved):
Everything ok. But I expect some merge conflicts with #292 (OP#2562)

Issue #285, opened 2023-07-27
OTA crashes at loading the tracks of the test video


arminkollascheck:
Loaded the newest version of OTA main. Installation and loading the start window worked fine. Tried to load the "Testvideo_Cars-Cyclist_FR20_2020-01-01_00-00-00.ottrk". After a second the loading occured and the image of Radeberg gets loaded. A sec later the windows and the cmd closes, so the program crashed? No error message occured in cmd.
![grafik](https://github.com/OpenTrafficCam/OTAnalytics/assets/73596011/b57f5fbe-d16b-4b3a-bca6-702210e7834f)

Same happens even if I enter a name and a date.
Python: 3.11.4
Recloned and Reinstalled OTA already.

No error or crash with "Sample_FR20_2020-01-01_00-00-00.ottrk", but obviously no picture shown.

briemla:
Thank you for the bug report. We will have a look at it. Given the terminal screenshot, I assume the error occurred on windows. Which version do you use? Windows 10 or 11?

arminkollascheck:
Windows 10

arminkollascheck:
Works now with the newest version.

PR #299, opened 2023-08-09
bug/2599-gui-crashes-before-completely-loading-tracks-on-windows

martinbaerwolff:
OP#2599

Re-adds updating a message in a popup in the viewmodel´s intersect_tracks_with_sections method (partly reverts #278 and OP##2538)


randy-seng (approved):

martinbaerwolff:
Fixes #285

PR #309, opened 2023-08-16
Feature/2581 use shapely intersects method to highlight tracks intersecting sections

randy-seng:
OP#2581
OP#2654


briemla (approved):
Looks good.

PR #310, opened 2023-08-17
Feature/2075 log and show every exception at top level

randy-seng:
OP#2075


briemla (commented):
Looks good so far. In my case the ".log" folder ist created inside the "OTAnalytics" folder. Outside of that folder would be better

briemla (changes requested):
Please rename the logging module to `logger` it shadows the python logging module.

briemla (approved):

PR #318, opened 2023-08-25
Feature/2452 clear section and flow repository when loading an otflow file

randy-seng:
* extract loading an otflow as a separate use case
* remove `load_otflow` from `Datastore`
* check that section id and name are unique before adding them to repository
* check that flow id and name are unique before adding them to repository
* warn user that currently created sections and flows will be deleted when loading otflow.

OP#2452

### DIFF
--- a/tests/OTAnalytics/application/use_cases/test_load_otflow.py
+++ b/tests/OTAnalytics/application/use_cases/test_load_otflow.py
@@ -1,0 +1,133 @@
+from pathlib import Path
+from typing import TypedDict
+from unittest.mock import Mock, call
+
+import pytest
+
+from OTAnalytics.application.datastore import FlowParser
+from OTAnalytics.application.use_cases.event_repository import ClearEventRepository
+from OTAnalytics.application.use_cases.flow_repository import (
+    AddFlow,
+    ClearFlows,
+    FlowAlreadyExists,
+)
+from OTAnalytics.application.use_cases.load_otflow import (
+    LoadOtflow,
+    UnableToLoadFlowFile,
+)
+from OTAnalytics.application.use_cases.section_repository import (
+    AddSection,
+    ClearSections,
+    SectionAlreadyExists,
+)
+from OTAnalytics.domain.flow import Flow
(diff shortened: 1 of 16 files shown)


tests/OTAnalytics/application/use_cases/test_load_otflow.py
```diff
@@ -0,0 +1,133 @@
...
+        flow_parser = Mock(spec=FlowParser)
+        flow_parser.parse.return_value = (
+            [mock_first_section, mock_second_section],
+            [mock_flow],
+        )
+
+        add_section = Mock(spec=AddSection)
+        add_flow = Mock(spec=AddFlow)
+        return {
+            "clear_all_sections": clear_all_sections,
```
briemla, line 66: The name strings of the parameters are used multiple times. Can we use constants instead of repeated strings?

tests/OTAnalytics/application/use_cases/test_section_repository.py
```diff
@@ -66,11 +69,12 @@ def test_add_section_with_different_names(
     ) -> None:
         use_case = AddSection(section_repository)
 
-        use_case.add(section_south)
+        use_case.__call__(section_south)
```
briemla, line 72: Can't we directly call `use_case(section_south)`?

briemla (changes requested):
Looks good. Only cosmetic changes requested.

tests/OTAnalytics/application/use_cases/test_load_otflow.py
randy-seng: Oh, this is because I declared the `MockDependencies` as a `TypedDict`. The python language server will be able to to give auto suggestion on possible key entries, which are String literals. 

Using constants is possible. In the case of `"clear_all_sections"`, the constant needs to be declared as

```python
CLEAR_ALL_SECTIONS: Literal["clear_all_sections"] = "clear_all_sections"
```
But this is a bit verbose.

KueblerJelle (approved):

PR #319, opened 2023-08-30
Feature/2649 restart with a fresh project without restarting the gui

randy-seng:
OP#2649


briemla (approved):
Looks good after changes of #318 have been checked.

PR #332, opened 2023-09-15
Task/2580 Limit progress updates

briemla:
OP#2580

### DIFF
--- a/OTAnalytics/application/progress.py
+++ b/OTAnalytics/application/progress.py
@@ -23,24 +23,34 @@ class AutoIncrementingProgressbar(Progressbar):
         sequence: Sequence,
         counter: Counter,
         notify: Optional[Callable[[], None]] = None,
+        step_percentage: int = 1,
     ) -> None:
         self._sequence = sequence
         self._counter = counter
         self._notify = notify
+        self._step_percentage = step_percentage * 100
         self._iterator = iter(self._sequence)
         self._counter.reset()
 
     def __iter__(self) -> Iterator:
+        total_elements = len(self._sequence)
+        step_size = self.__get_step_size(total_elements)
         self._counter.reset()
         self._iterator = iter(self._sequence)
         while True:
             try:
                 yield next(self._iterator)
                 self._counter.increment(1)
-                if self._notify:
-                    self._notify()
+                counter_value = self._counter.get_value()
+                if counter_value == total_elements or counter_value % step_size == 0:
+                    if self._notify:
+                        self._notify()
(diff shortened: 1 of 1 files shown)


OTAnalytics/application/progress.py
```diff
@@ -23,25 +23,35 @@ def __init__(
         sequence: Sequence,
         counter: Counter,
         notify: Optional[Callable[[], None]] = None,
+        step_percentage: int = 1,
```
randy-seng, line 26: Since `step_percentage` is of type int, I assume that the values are within in the bounds of [1,100]. Is that correct? I think it would be best to validate the value  in the constructor.

OTAnalytics/application/progress.py
```diff
@@ -23,25 +23,35 @@ def __init__(
...
     ) -> None:
         self._sequence = sequence
         self._counter = counter
         self._notify = notify
+        self._step_percentage = step_percentage * 100
```
randy-seng, line 31: Why do we need to multiply the value by 100?

OTAnalytics/application/progress.py
```diff
@@ -23,25 +23,35 @@ def __init__(
...
-                    self._notify()
+                counter_value = self._counter.get_value()
+                if counter_value == total_elements or counter_value % step_size == 0:
+                    if self._notify:
+                        self._notify()
             except StopIteration:
                 return
 
+    def __get_step_size(self, total_elements: int) -> int:
+        step_size = int(total_elements / self._step_percentage)
```
randy-seng, line 52: This is related to comment above. If `_step_percentage` is within the bounds of [1,100] and 100 is not multiplied to it, shouldn't  `step_size` be calculated as follows: `(total_elements * self._step_percentage) / 100`?

randy-seng (changes requested):
I have some questions and a failed unit tests needs some fixing~

OTAnalytics/application/progress.py
briemla: Yes, I added a validation.

OTAnalytics/application/progress.py
briemla: I do not know anymore.

OTAnalytics/application/progress.py
briemla: I fixed the calculation. The old one was simply wrong.

randy-seng (approved):
Looks good! I set the `step_percentage` to 5% to use the new option.

PR #355, opened 2023-10-05
Feature/2929 take each possible class into account even if class isnt assigned to any track

randy-seng:
OP#2929


tests/OTAnalytics/plugin_parser/test_otvision_parser.py
```diff
@@ -261,11 +261,11 @@ def test_parse_ottrk_sample(
         ottrk_data = track_builder_setup_with_sample_data.build_ottrk()
         ottrk_file = test_data_tmp_dir / "sample_file.ottrk"
         _write_bz2(ottrk_data, ottrk_file)
-        result_tracks = ottrk_parser.parse(ottrk_file)
+        parse_result = ottrk_parser.parse(ottrk_file)
 
         expected_track = track_builder_setup_with_sample_data.build_track()
 
-        assert result_tracks == PythonTrackDataset.from_list([expected_track])
+        assert parse_result.tracks == PythonTrackDataset.from_list([expected_track])
```
briemla, line 268: Can we also assert the detection classes?

PR #356, opened 2023-10-05
feature/3227-export-event-list-as-csv-using-cli-(choose-export-format-on-cli)

randy-seng:
OP#3227

### DIFF
--- a/OTAnalytics/plugin_ui/cli.py
+++ b/OTAnalytics/plugin_ui/cli.py
@@ -1,5 +1,6 @@
 from argparse import ArgumentParser
 from dataclasses import dataclass
+from enum import Enum
 from pathlib import Path
 from typing import Iterable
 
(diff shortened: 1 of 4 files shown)


OTAnalytics/plugin_ui/cli.py
```diff
@@ -59,6 +72,8 @@ class CliArguments:
     track_files: list[str]
     sections_file: str
     eventlist_filename: str
+    event_format: EventFormat
```
briemla, line 75: Can the event_format be of type `EventListExporter`? See comments below.

OTAnalytics/plugin_ui/cli.py
```diff
@@ -120,9 +152,24 @@ def parse(self) -> CliArguments:
...
 
+    def _parse_event_format(self, event_format: str) -> EventFormat:
+        match event_format.lower():
+            case EventFormat.CSV.value:
+                return EventFormat.CSV
+            case EventFormat.EXCEL.value:
+                return EventFormat.EXCEL
+            case _:
+                return EventFormat.OTEVENTS
+
```
briemla, line 173: If we parse the input and directly return the required `EventlistExporter` we do not need the `match ... case ...` statement when exporting.

OTAnalytics/plugin_ui/cli.py
```diff
@@ -342,6 +385,25 @@ def _get_sections_file(file: str) -> Path:
...
+                exporter.export(events, sections, actual_save_path)
+            case EventFormat.EXCEL:
+                exporter = self._event_list_export_formats[OTC_EXCEL_FORMAT_NAME]
+                actual_save_path = save_path.with_suffix(f".{EventFormat.EXCEL.value}")
+                exporter.export(events, sections, actual_save_path)
+            case _:
+                exporter = self._event_list_export_formats[OTC_OTEVENTS_FORMAT_NAME]
+                actual_save_path = save_path.with_suffix(
+                    f".{EventFormat.OTEVENTS.value}"
+                )
```
briemla, line 403: This is three times nearly the same code. If we save an instance of `EventlistExporter` in CliArguments, we can directly use the instance here and get rid of the `match ... case ...`.

briemla (changes requested):
Looks good so far. I just have a bunch of questions all about the same topic.

PR #358, opened 2023-10-10
Feature/2889 show info that offset of currently selected section and visualized offset are different

randy-seng:
OP#2889


briemla (changes requested):
The color is not updated on:
- changes of the offset value of the currently selected section
- changes of the currently visualized offset

The color is only updated on selection changes.

briemla (approved):

PR #365, opened 2023-10-12
Bug/3237 creating legend with loc=best is possibly slowing down track visualization

randy-seng:
OP#3237

### DIFF
--- a/OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
+++ b/OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
@@ -504,7 +504,6 @@ class TrackGeometryPlotter(MatplotlibPlotterImplementation):
 
         Args:
             track_df (DataFrame): tracks to plot
-            alpha (float): transparency of the lines
             axes (Axes): axes to plot on
         """
         seaborn.lineplot(
@@ -519,8 +518,16 @@ class TrackGeometryPlotter(MatplotlibPlotterImplementation):
             alpha=self._alpha,
             ax=axes,
             palette=self._color_palette_provider.get(),
-            legend=self._enable_legend,
         )
+        if self._enable_legend:
+            print("creating legend")
+            legend = axes.legend(loc="upper right")
+            legend.set_alpha(1)
+        else:
+            # Somehow enabling a legend from previous call to this method persists.
+            # Need to manually remove it, if legend exists.
+            if existing_legend := axes.legend_:
+                existing_legend.remove()
 
 
 class TrackStartEndPointPlotter(MatplotlibPlotterImplementation):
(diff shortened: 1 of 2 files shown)


OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
```diff
@@ -519,8 +518,16 @@ def _plot_dataframe(self, track_df: DataFrame, axes: Axes) -> None:
             alpha=self._alpha,
             ax=axes,
             palette=self._color_palette_provider.get(),
-            legend=self._enable_legend,
         )
+        if self._enable_legend:
+            print("creating legend")
```
martinbaerwolff, line 523: Can this print be removed?

OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
```diff
@@ -519,8 +518,16 @@ def _plot_dataframe(self, track_df: DataFrame, axes: Axes) -> None:
             alpha=self._alpha,
             ax=axes,
             palette=self._color_palette_provider.get(),
-            legend=self._enable_legend,
         )
+        if self._enable_legend:
```
martinbaerwolff, line 522: Where did self._enable_legend come from? Is this for a future feature to toggle the legend?

OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
```diff
@@ -519,8 +518,16 @@ def _plot_dataframe(self, track_df: DataFrame, axes: Axes) -> None:
...
+            print("creating legend")
+            legend = axes.legend(loc="upper right")
+            legend.set_alpha(1)
+        else:
+            # Somehow enabling a legend from previous call to this method persists.
```
martinbaerwolff, line 527: Is this (maybe) because the axes persist, too?

martinbaerwolff (changes requested):
Looks good, just one small thing and two questions just for information.

OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
randy-seng: I just had a look at the code. The `track_viz.MatplotlibTrackPlotter.plot`  method creates the axes via `_create_figure` and `_create_axes` methods. Nothing indicates that existing axes are being used, at least to me 🤔

OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
randy-seng: `_enable_legend` is a member variable of `TrackGeometryPlotter`. In `main_application.py` several instances of `TrackGeometryPlotter` are created and passed to several layers that want to plot tracks on the canvas. Thus, we don't differentiate between plotters used in the different layers using it. Hence, the need to for `_enable_legend`. If you have a look at the `_create_layers` method in `main_application.py` you can see how plotters are being configured.

In the future, the member variable will be not necessary anymore, if the legend is displayed outside of the canvas.

OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
martinbaerwolff: Hm ok, anyway, we have this in mind in case of plotting issues :)

martinbaerwolff (approved):

PR #366, opened 2023-10-13
feature/3248-speed-up-loading-of-frames-in-otconfigotanalytics

briemla:
OP#3248


martinbaerwolff (approved):
Is quite fast now - loading first frames of 24h (96 videos) from SSD took approx. 3.9 seconds on Windows (Intel Core i9) and approx. 2 sec on Mac (M1 Pro). Thanks!

PR #367, opened 2023-10-13
feature/3248-speed-up-loading-of-frames-in-otconfigotanalytics-2

briemla:
OP#3248


martinbaerwolff (approved):
Feels good! Perceived difference in display time between cached and non-cached frames is imperceptible when loading from SSD and minimal when loading from poorly connected network drive.

PR #370, opened 2023-10-16
Bug/3061 inspect possible bug with cutting section

randy-seng:
OP#3061


briemla (approved):
Looks good. I could not see any tracks missing after cutting.

PR #380, opened 2023-10-23
Feature/3266 calculate geometry once per used offset group section by offset before creating intersection objects

randy-seng:
OP#3266


tests/OTAnalytics/domain/test_event.py
```diff
@@ -26,8 +26,9 @@
 )
 from OTAnalytics.domain.geometry import DirectionVector2D, ImageCoordinate
 from OTAnalytics.domain.section import SectionId
-from OTAnalytics.domain.track import Detection, PythonDetection, TrackId
+from OTAnalytics.domain.track import Detection, TrackId
 from OTAnalytics.domain.types import EventType, EventTypeParseError
+from OTAnalytics.plugin_datastore.python_track_store import PythonDetection
```
briemla, line 31: This is a reference from the domain module into a plugin module. We could easily mock the usage in `valid_detection` to remove this dependency.

OTAnalytics/plugin_intersect/shapely/create_intersection_events.py
```diff
@@ -0,0 +1,307 @@
...
+        tracks: Iterable[Track],
+        section: Section,
+        event_builder: EventBuilder,
+    ) -> list[Event]:
+        events: list[Event] = []
+        section_geometry: LineString = self._geometry_builder.create_section(section)
+        for track in tracks:
+            track_geometry = self._track_table.look_up(track)
+            if not track_geometry.intersects(section_geometry):
+                continue
```
briemla, line 127: After merging OP#3493. We can reuse the intersection_repository to speed up event creation.

OTAnalytics/plugin_intersect/shapely/create_intersection_events.py
```diff
@@ -0,0 +1,307 @@
...
+            event_builder=event_builder,
+        )
+        events.extend(create_intersection_events.create())
+    return events
+
+
+def group_sections_by_offset(
+    sections: Iterable[Section],
+) -> Mapping[RelativeOffsetCoordinate, Iterable[Section]]:
+    grouped_sections: dict[RelativeOffsetCoordinate, list[Section]] = {}
```
briemla, line 300: We can use `defaultdict(list)` here. This will remove the if else in the loop.

briemla (changes requested):
looks nice so far. Just a few comments. I fixed some smaller things.

OTAnalytics/plugin_intersect/shapely/create_intersection_events.py
randy-seng: Sounds good! I'll implement the re-use of the `intersection_repository` once  #385 has been merged~

PR #396, opened 2023-11-07
Feature/3531 add cli option to specify log file path

randy-seng:
OP#3531

### DIFF
--- a/OTAnalytics/plugin_ui/cli.py
+++ b/OTAnalytics/plugin_ui/cli.py
@@ -335,11 +335,13 @@ class OTAnalyticsCli:
         save_stem = cli_args.save_name if cli_args.save_name else otflow.stem
         save_suffix = f"_{cli_args.save_suffix}" if cli_args.save_suffix else ""
 
-        if (
-            (valid_path := Path(cli_args.save_name)).parent
-        ).is_absolute() or valid_path.is_symlink():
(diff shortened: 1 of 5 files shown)


OTAnalytics/plugin_ui/cli.py
```diff
@@ -318,11 +335,13 @@ def _create_save_path(self) -> Path:
...
-            (valid_path := Path(cli_args.save_name)).parent
-        ).is_absolute() or valid_path.is_symlink():
-            return valid_path.with_name(valid_path.stem + save_suffix)
-        return otflow.with_name(save_stem + save_suffix)
+        if not self.cli_args.save_name:
+            # No save name specified. Take otflow name as stem for save path.
+            return otflow.with_name(save_stem + save_suffix)
+
+        # Save name is either absolute or relative path.
+        save_path = Path(self.cli_args.save_name).expanduser()
```
briemla, line 343: Do we also need to call `resolve`?

randy-seng: Actually it is not required to use `resolve` at all. The same goes for all other `resolve` usages in `cli.py`. `pathlib` will be able to resolve relative paths regardlessly with or without the call to `resolve`. I removed all usages of `resolve` in my newest commit.

briemla (approved):

PR #411, opened 2023-11-21
bug/3670-background-image-is-not-updated-when-video-is-added

briemla:
OP#3670


randy-seng (changes requested):
Looks good! But shouldn't we delete the old `TrackBackgroundPlotter` in `OTAnalytics.plugin_ui.visualization` including all related tests?

randy-seng (approved):

PR #432, opened 2024-01-19
Feature/3995 pandas implementation to cut tracks with sections

randy-seng:
OP#3995
OP#3996
OP#4063


briemla (approved):
Nice one. +1 for using index manipulation to speed up cutting :-)

briemla (approved):

PR #435, opened 2024-01-23
bug/4094-count-export-fails-on-empty-dataframe

briemla:
OP#4094


OTAnalytics/plugin_parser/export.py
```diff
@@ -35,11 +35,15 @@ def __init__(self, output_file: str) -> None:
     def export(self, counts: Count) -> None:
         logger().info(f"Exporting counts to {self._output_file}")
         dataframe = self.__create_data_frame(counts)
+        if dataframe.empty:
```
randy-seng, line 38: Can we add a test case for this?~

randy-seng (changes requested):
Can we add a test case for this bug?~

randy-seng (approved):

PR #474, opened 2024-03-05
Add run configurations for scalene, cli and prefilled gui using test and benchmark data

briemla:


randy-seng (commented):
Looks good. But do we need to add scalene to the requirements-dev.txt? But I remember adding scalene to the requirements-dev.txt caused some problems on the GitHub CI runner.

randy-seng (approved):

PR #478, opened 2024-03-08
task/4602-export-tracks-as-csv

briemla:
OP#4602


OTAnalytics/plugin_parser/track_export.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def export(self, specification: TrackExportSpecification) -> None:
+        dataframe = self._get_data()
+        dataframe = self._set_column_order(dataframe)
+        output_path = specification.save_path.with_suffix(".tracks.csv")
+        dataframe.to_csv(output_path, index=False)
+
+    def _get_data(self) -> DataFrame:
+        dataset = self._track_repository.get_all()
+        if isinstance(dataset, PandasTrackDataset):
```
randy-seng, line 27: We could also do an `isinstance` on `PandasDataframeProvider`. The `isinstance` on `FilteredPandasTrackDataset` can then be omitted.

OTAnalytics/plugin_parser/track_export.py
```diff
@@ -0,0 +1,58 @@
...
+)
+from OTAnalytics.domain import track
+from OTAnalytics.domain.track_repository import TrackRepository
+from OTAnalytics.plugin_datastore.track_store import (
+    FilteredPandasTrackDataset,
+    PandasTrackDataset,
+)
+
+
+class CsvTrackExport(ExportTracks):
```
randy-seng, line 15: Do we need to write a test case for this class?

randy-seng (changes requested):
Looks good! I just have a couple of questions/comments~

OTAnalytics/plugin_parser/track_export.py
briemla: It was the wrong import. I move the PandasDataFrameProvider into the pandas module to be more consistent and to prevent circular dependencies.

OTAnalytics/plugin_parser/track_export.py
briemla: Done.

randy-seng (approved):

PR #484, opened 2024-03-13
feature/4643-move-time-filter-to-previous-or-next-event

briemla:
OP#4643


randy-seng (changes requested):
Looks good!! Just a few comments

randy-seng (approved):

PR #485, opened 2024-03-15
task/4190-implement-regression-test-for-otanalytics-cli-using-test-data

briemla:
OP#4190
OP#4146

### DIFF
--- a/tests/regression_otanalytics.py
+++ b/tests/regression_otanalytics.py
@@ -2,7 +2,7 @@ from pathlib import Path
 
 import pytest
 
-from OTAnalytics.application.datastore import FlowParser
+from OTAnalytics.application.parser.flow_parser import FlowParser
 from OTAnalytics.plugin_parser.otvision_parser import OtFlowParser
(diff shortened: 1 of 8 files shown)


tests/regression_otanalytics.py
```diff
@@ -0,0 +1,155 @@
...
+        expected_counts_file = (
+            Path(test_data_dir / save_name)
+            .with_suffix(f".counts_{count_interval}min.csv")
+            .absolute()
+        )
+        with open(actual_counts_file, mode="r") as actual:
+            actual_lines = sorted(actual.readlines())
+            with open(expected_counts_file, mode="r") as expected:
+                expected_lines = sorted(expected.readlines())
+                assert actual_lines == expected_lines
```
randy-seng, line 155: We could extract this as a separate function since it is also  used in l.135

randy-seng (changes requested):
Lookds good! Just two comments that I have~

randy-seng (approved):

PR #486, opened 2024-03-15
Feature/4641 implement save and save as

randy-seng:
OP#4641
OP#4642


tests/OTAnalytics/application/use_cases/test_load_otconfig.py
```diff
@@ -0,0 +1,92 @@
...
+
+        otconfig = Mock()
+        otconfig.project = project
+        otconfig.videos = Mock()
+        otconfig.sections = Mock()
+        otconfig.flows = Mock()
+        return otconfig
+
+    def test_load(self, otconfig: OtConfig) -> None:
+        clear_repositories = Mock()
```
briemla, line 30: Should we test that `clear_repositories` is called?

tests/OTAnalytics/application/use_cases/test_load_otconfig.py
```diff
@@ -0,0 +1,92 @@
...
+            Mock(),
+            Mock(),
+        )
+        observer = Mock()
+        load_otconfig.register(observer)
+        file = Mock()
+        with pytest.raises(UnableToLoadOtconfigFile):
+            load_otconfig.load(file)
+
+        assert clear_repositories.call_count == 2
```
briemla, line 90: Why is the repository cleared twice?

randy-seng: A `SectionAlreadyExists` or `FlowAlreadyExists` Exception will be thrown if  a section or flow already exists. This implies that the OTFlow or OTConfig file is not valid.
Since we are sequentially adding each section or flow to the repository, the repository could be only partly loaded when any of the above exceptions occur. To avoid this state, we decided to just clear the repositories again.

The correct behaviour would be to restore the state before a configuration file was loaded if an error occurs.
This would require validation logic in the `AddSections` or `AddFlows` use cases. But maybe that is actually the responsibility of the `FlowParser` or `ConfigFileParser` to return a validly parsed configuration. I leaning to the latter solution. @briemla

briemla: The latter solution sounds good. Can you please file a feature / epic in OP?

briemla (approved):

PR #488, opened 2024-03-21
bug/4709-activation-of-filter-by-date-is-broken

briemla:
OP#4709


randy-seng (changes requested):
After loading a track file and activating the check box of the datetime filter, you are not able to click on the "Filter By Date" button since it is still deactivated. Only after clicking any of the Video Control buttons it is enabled.

randy-seng (approved):

randy-seng (approved):
Looks good! I have fixed the failed unit tests. Also the DefaultFilterRange wasn't set when switching to the next event at the first occasion. I have reassigned this PR to you so that you can have a look at my changes. You can go ahead and merge if everything is good.

PR #489, opened 2024-03-22
Feature/2056 update metadata in pyprojecttoml

randy-seng:
OP#2056
OP#4686


.github/workflows/publish-testpypi.yml
```diff
@@ -0,0 +1,79 @@
...
+        uses: JesseTG/rm@v1.0.3
+        with:
+          path: dist/
+      - name: Install pypa/build
+        run: >-
+          python -m
+          pip install
+          build
+          --user
+      - name: Build a binary wheel and a source tarball
```
frunika, line 65: Do we need to build os dependend wheels? At least mac os arm and x64 need to be different, don't they?

pyproject.toml
```diff
@@ -3,16 +3,41 @@ requires = ["hatchling", "hatch-requirements-txt"]
 build-backend = "hatchling.build"
 
 [project]
-name = "OTAnalytics"
-version = "0.0.1"
-dynamic = ["dependencies"]
+name = "OTAnalytics-EXAMPLE-TEST-DISTRIBUTION"
+dynamic = ["dependencies", "version"]
+authors = [
```
frunika, line 8: Should we use a generic `{ name="OpenTrafficCam Contributors", email="team@opentrafficcam.org" }`?

.github/workflows/publish-testpypi.yml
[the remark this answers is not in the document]
briemla: Yes, this is part of OP#2132. The code of this part will be moved to our reusable workflows. It was created here to test the functionality with OTAnalytics with faster feedback cycles.

.github/workflows/publish-testpypi.yml
briemla: > `python3 -m build --wheel`
> The `wheel` package will detect that the code is not pure Python, and build a wheel that’s named such that it’s only usable on the platform that it was built on. For details on the naming of wheel files, see [PEP 425](https://peps.python.org/pep-0425/).

The Python build module will take care of that one. In our case, we only build a pure python wheel, because we do not use platform specific extensions ourself. Our dependencies will use platform specific extension which will be installed by pip.

.github/workflows/publish-testpypi.yml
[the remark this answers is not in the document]
randy-seng: Since the workflow file has been tested should we remove it from here and add it as an attachment to OP#2132?

briemla (dismissed):

PR #490, opened 2024-03-26
bug/4735-event-file-is-bigger-after-updating-event-creation

briemla:
OP#4735

### DIFF
--- a/OTAnalytics/domain/event.py
+++ b/OTAnalytics/domain/event.py
@@ -376,6 +376,7 @@ class EventRepository:
         """
         self.__do_add(event)
         self.__sort()
+        self.__discard_duplicates()
         self._subject.notify(EventRepositoryEvent([event], []))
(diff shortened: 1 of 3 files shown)


randy-seng (commented):
The event creation has slowed down, I added some changes to speed up removing duplicate events from the event repository. It should be faster now. The benchmark is also running as of writing this comment. @briemla Can you review my changes. I'll approve this PR when I get your ok.

briemla:
> The event creation has slowed down, I added some changes to speed up removing duplicate events from the event repository. It should be faster now. The benchmark is also running as of writing this comment. @briemla Can you review my changes. I'll approve this PR when I get your ok.

Looks good to me. I switched the invocation order in `add`. Duplicates will now be discarded before sorting as it already was done in `add_all`. This is to be consistent between calls.

PR #506, opened 2024-04-19
bug/4915-moving-to-the-next-video-using-the-date-range-filter-or-the-video-control-does-not-work

randy-seng:
OP#4915


briemla (changes requested):
Looks good. Just a few small recommendations.

briemla (approved):

PR #514, opened 2024-04-30
bug/4710-bbox-in-miovision-videos-are-completly-out-of-sync-with-background-frame

briemla:
OP#4710

### DIFF
--- a/OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
+++ b/OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
@@ -57,7 +57,7 @@ from OTAnalytics.plugin_datastore.track_store import PandasDataFrameProvider
 from OTAnalytics.plugin_filter.dataframe_filter import DataFrameFilterBuilder
 
 """Frames start with 1 in OTVision but frames of videos are loaded zero based."""
-FRAME_OFFSET = 1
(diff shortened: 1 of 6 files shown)


randy-seng (changes requested):
oks good! I found some print statements. Should we log them?

OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
```diff
@@ -57,7 +57,7 @@
 from OTAnalytics.plugin_filter.dataframe_filter import DataFrameFilterBuilder
 
 """Frames start with 1 in OTVision but frames of videos are loaded zero based."""
-FRAME_OFFSET = 1
+FRAME_OFFSET = 0
```
randy-seng, line 61: Has this changed?

randy-seng (commented):
I have one more question~

OTAnalytics/plugin_prototypes/track_visualization/track_viz.py
briemla: Yes, because we use the frame information to filter the tracks and to find the matching background image. This offset is a relict of the old strategy.

randy-seng (approved):

PR #515, opened 2024-04-30
Feature/4995 fallback to video file name if relative path can not be found

randy-seng:
#509 needs to be merged first

OP#4995


briemla (approved):
Looks good.

PR #516, opened 2024-05-02
bug/missing-metadata-when-loading-otconfig-via-cli

briemla:


OTAnalytics/application/use_cases/preload_input_files.py
```diff
@@ -8,17 +9,23 @@ class PreloadInputFiles:
...
         load_otflow: LoadOtflow,
         apply_cli_cuts: ApplyCliCuts,
     ):
         self._load_track_files = load_track_files
+        self._load_otconfig = load_otconfig
         self._load_otflow = load_otflow
         self._apply_cli_cuts = apply_cli_cuts
 
     def load(self, run_config: RunConfiguration) -> None:
+        if run_config.config_file:
```
randy-seng, line 22: I added a fix to prevent loading the track files twice. I am not sure what behavior we want to adapt. But this feels right to me. Specifying the otconfig file should take precedence over the  otflow file. 

The same goes for  the cli option `--ottrks`, in which the track files specified within the OTConfig should be loaded. Whereas the track files specified via `--ottrks` should be ignored. The `TrackRepository` does ensure the uniqueness of tracks though. But loading the same file twice will raise an error in the class `VideosMetadata` since the video metadata with the same start date already exists. 

Any thoughts on this? @briemla

randy-seng (commented):
I fixed the failed unit test and I also found another bug where the track files are being loaded twice. See the comments down below for more details.

randy-seng (approved):

PR #527, opened 2024-06-07
Feature/4906 Add flag to show svz tab

briemla:
OP#4906

### DIFF
--- a/OTAnalytics/plugin_parser/argparse_cli_parser.py
+++ b/OTAnalytics/plugin_parser/argparse_cli_parser.py
@@ -28,6 +28,12 @@ class ArgparseCliParser(CliParser):
             required=False,
         )
         self._parser.add_argument(
+            "--show_svz",
+            action="store_true",
+            help="Show SVZ-Tab in OTAnalytics GUI. If omitted the tab will be hidden.",
+            required=False,
+        )
+        self._parser.add_argument(
             "--config",
             type=str,
             help="Path to otconfig file.",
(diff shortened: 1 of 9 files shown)


OTAnalytics/plugin_ui/customtkinter_gui/frame_project.py
```diff
@@ -73,6 +73,8 @@ def _get_widgets(self) -> None:
         self.frame_svz_metadata = FrameSvzMetadata(
             master=self.tab(self._svz_title), viewmodel=self._viewmodel
         )
+        if not self._viewmodel.show_svz():
+            self.delete(self._svz_title)
```
briemla, line 77: The svz tab and all ui elements are created. Afterwards, the tab is deleted to be hidden. This was necessary, because our current approach of registering frames to the view model raises an exception if one of the frames is missing. I did not want to change that. This will be changed on the transition to a web ui.

OTAnalytics/plugin_parser/argparse_cli_parser.py
```diff
@@ -27,6 +27,12 @@ def _setup(self) -> None:
             help="Start OTAnalytics CLI. If omitted OTAnalytics GUI will be started.",
             required=False,
         )
+        self._parser.add_argument(
+            "--show_svz",
```
randy-seng, line 31: Can we rename it to `--show-svz` so that this argument is uniform to the other arguments?

randy-seng (changes requested):
Looks good! Just a tiny change request :)

OTAnalytics/plugin_parser/argparse_cli_parser.py
briemla: Sure, I already changed it.

randy-seng (approved):

PR #540, opened 2024-07-10
bug/5341-visualization-shows-incorrect-number-of-tracks

briemla:
OP#5341


randy-seng (changes requested):
Looks good and works! I have a couple of small change requests

randy-seng (approved):

PR #550, opened 2024-08-20
Bug/5669 replace types all in mypy in pre commit configyaml

randy-seng:
OP#5669


OTAnalytics/plugin_datastore/track_store.py
```diff
@@ -670,18 +678,11 @@ def _get_dataset_with_classes(self, classes: list[str]) -> PandasTrackDataset:
...
-            tracks_to_remove
-        )
         return PandasTrackDataset(
-            self._other.track_geometry_factory,
-            filtered_df,
-            updated_geometry_datasets,
-            self._other.calculator,
+            track_geometry_factory=self._other.track_geometry_factory,
+            dataset=filtered_df,
+            geometry_datasets=None,
```
briemla, line 684: I know we discussed that `geometry_datasets` should be `None` here. Please document why.

OTAnalytics/plugin_parser/track_export.py
```diff
@@ -43,11 +43,14 @@ def _get_data(self) -> DataFrame:
...
-            [detection.to_dict() for detection in track.detections]
-            for track in dataset.as_list()
-        ]
-        return DataFrame.from_dict(detections)
+        detections = []
+        for _track in dataset.as_list():
+            track_classification = _track.classification
+            for detection in _track.detections:
+                current = detection.to_dict()
+                current[track.TRACK_CLASSIFICATION] = track_classification
```
briemla, line 51: Why do we need to change the track_classification for each detection?
Is it incorrect in the dataframe?

briemla (commented):
Looks good. Only two comments.

OTAnalytics/plugin_parser/track_export.py
randy-seng: We are actually not changing `track_classification`, but adding it to the detection dictionary, since a `Detection` object does not have a member called track_classification. The information about the `track_classification` is only contained in the `Track` class.

OTAnalytics/plugin_datastore/track_store.py
randy-seng: I added a comment in the code. Please check if the explanation is acceptable.

briemla (approved):

PR #551, opened 2024-08-22
feature 2056 update metadata in pyprojecttoml

frunika:
OP#2056
OP#5577


randy-seng (approved):
I also specified only the files we need to download in the regression test and benchmark. You can merge it if everything is good.

PR #556, opened 2024-10-01
Bug/5938 indexerror list index out of range when updating flow highlighting in otanalytics app

briemla:
OP#5938


randy-seng (approved):
Looks good!

PR #558, opened 2024-10-11
Task/5990 show track statistics

schubsen:
OP#5990


OTAnalytics/application/use_cases/highlight_intersections.py
```diff
@@ -59,6 +63,176 @@ def get_ids(self) -> set[TrackId]:
...
+        intersection_repository: IntersectionRepository,
+    ) -> None:
+        self._get_cutting_sections = get_cutting_sections
+        self._get_all_sections = get_all_sections
+        self._tracks_intersecting_sections = tracks_intersecting_sections
+        self._get_section_by_id = get_section_by_id
+        self._intersection_repository = intersection_repository
+
+    def get_ids(self) -> set[TrackId]:
+        return TracksIntersectingGivenSections(
```
briemla, line 91: Instead of `TracksIntersectingGivenSections` a new use case is necessary. Because `TracksIntersectingGivienSections` calculates line intersections. No track has a line intersection with a cutting section.
The new use case should be similar to `SimpleTracksIntersectingSections` in `OTAnalytics/plugin_intersect/simple_intersect.py`. Instead of using `intersecting_tracks` of `TrackDataset` in line 25, use `contained_by_sections`. This will check if detections are inside the area of the given sections.

briemla (changes requested):

OTAnalytics/application/use_cases/track_statistics.py
```diff
@@ -40,29 +41,31 @@ def __init__(
...
+        ids_inside_cutting_sections = self._track_ids_inside_cutting_sections()
+
+        track_count_inside = len(ids_inside_cutting_sections)
+        track_count = len(ids_all)
+
+        track_count_outside = track_count - track_count_inside
 
         track_count_inside_not_intersecting = len(
-            self._inside_cutting_sections.get_ids().difference(
+            ids_inside_cutting_sections.difference(
```
briemla: In a typical setup (all sections inside the cutting-section) the difference between two sets would be correct. However, it is possible to define a section outside the cutting section. This might result in more tracks intersecting a section than being inside the cutting section. A share greater 100% would indicate that. @MichHeilig should we show an indicator if section are located outside the current cutting section?

OTAnalytics/plugin_ui/customtkinter_gui/dummy_viewmodel.py
```diff
@@ -1850,5 +1850,9 @@ def set_frame_track_statistics(self, frame: AbstractFrameTrackStatistics) -> Non
         self._frame_track_statistics = frame
 
     def update_track_statistics(self, _: EventRepositoryEvent) -> None:
+        if self._frame_track_statistics is None:
+            raise MissingInjectedInstanceError(
+                type(self._frame_track_statistics).__name__
```
briemla, line 1855: This is not going to work as expected. This is equivalent to `type(None).__name__`. It might be a relict of one of the other frames where we missed to change the error message.

briemla (changes requested):

OTAnalytics/plugin_ui/customtkinter_gui/dummy_viewmodel.py
randy-seng: I refactored this introducing properties for each instance that might not be injected yet. Raising the `MissingInjectedInstanceError` is now handled in the respective properties.

randy-seng:
@briemla I put the track statistics, visualization filters, and video control UI elements to a scrollable frame.
Due to the addition of the track statistics frame, the video control element got cut off.

briemla (approved):

PR #564, opened 2024-11-15
Feature/6490 display the number of assigned tracks per flow in the flow table

randy-seng:
OP#6490
Requires #558 to be merged first

### DIFF
--- a/OTAnalytics/application/use_cases/flow_statistics.py
+++ b/OTAnalytics/application/use_cases/flow_statistics.py
@@ -1,3 +1,5 @@
+from collections import defaultdict
+
 from OTAnalytics.application.use_cases.get_road_user_assignments import (
     GetRoadUserAssignments,
 )
@@ -19,7 +21,7 @@ class NumberOfTracksAssignedToEachFlow:
         return result
 
     def _initialize_flow_track_counts(self) -> dict[FlowId, int]:
-        flows = {}
+        flows = defaultdict(lambda: 0)
         for flow in self._flow_repository.get_all():
             flows[flow.id] = 0
         return flows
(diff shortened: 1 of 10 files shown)


OTAnalytics/application/use_cases/flow_statistics.py
```diff
@@ -0,0 +1,25 @@
...
+
+class NumberOfTracksAssignedToEachFlow:
+    def __init__(
+        self, get_assignments: GetRoadUserAssignments, flow_repository: FlowRepository
+    ) -> None:
+        self._get_assignments = get_assignments
+        self._flow_repository = flow_repository
+
+    def get(self) -> dict[FlowId, int]:
+        result = self._tracks_assigned_to_flows()
```
briemla, line 15: Why do we not just use `defaultdict`?
If we need the method, please find a better name.

briemla (changes requested):

OTAnalytics/application/use_cases/flow_statistics.py
randy-seng: We could use a defaultdict. But we still need to iterate over all existing flows first, since there can flows that do not have any tracks assigned to. Populating the dictionary with all existing flows ensures that no flows are missing in the result.

briemla: We only need to add 0 for all flows, if we iterate over the returned dictionary or if we want to export it as a whole.  For the current usage by the ui, we request the number per flow-id by iterating over all flow-ids in the flow-repository, see `frame_flows.py` in line 136. If we request the count for a missing flow-id this will fail. 
As both method iterate over all flows in the flow repository, there is actually no problem. But it might change it the future. I therefore vote to use a `defaultdict` to lower the coupling between the ui class and the use case.

briemla (changes requested):

briemla (approved):

PR #585, opened 2024-12-19
bug/6594-only-first-frame-of-video-is-shown

randy-seng:
Introduced a fallback mechanism in `PyAvVideoReader` to retrieve the total frame count from `VideosMetadata` if not available in the video stream. Updated related tests and refactored initialization to ensure `VideosMetadata` is passed where required.

OP#6594


briemla (approved):
Works as expected!

PR #589, opened 2025-01-10
bug/2665-ensure-no-file-is-loaded-twice

randy-seng:
The pull request adds a check to skip already loaded track files, along with a warning log. It also refactors test cases to eliminate code duplication.

OP#2665

### DIFF
--- a/tests/OTAnalytics/application/use_cases/test_load_track_files.py
+++ b/tests/OTAnalytics/application/use_cases/test_load_track_files.py
@@ -1,12 +1,11 @@
-from datetime import datetime
+from dataclasses import dataclass
 from pathlib import Path
 from unittest.mock import MagicMock, Mock, call, patch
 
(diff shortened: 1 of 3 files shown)


tests/OTAnalytics/application/use_cases/test_load_track_files.py
```diff
@@ -1,89 +1,151 @@
...
+            track_ids=[],
+            video_files=[],
+            track_files=[some_file],
+            existing_track_files=[some_file],
+            classes=set(),
+        )
+        target = create_target(given)
+
+        target([some_file])
+        mock_load.assert_not_called()
```
schubsen, line 47: a test loading two files where one of them exists would be useful

schubsen (changes requested):

schubsen (approved):

PR #619, opened 2025-02-06
feature/7185-refactor-use-case-to-add-section-with-its-geometry

briemla:
OP#7246


OTAnalytics/application/use_cases/editor/section_editor.py
```diff
@@ -0,0 +1,138 @@
...
+        raise MissingCoordinate("First coordinate is missing")
+    elif len(coordinates) == 1:
+        raise MissingCoordinate("Second coordinate is missing")
+
+
+def validate_section_information(
+    meta_data: dict, coordinates: list[tuple[int, int]]
+) -> None:
+    validate_coordinates(coordinates)
+    if not meta_data:
```
randy-seng, line 39: `meta_data`s type signature should be denoted as `dict | None` if it can be `None`.

randy-seng (changes requested):

OTAnalytics/application/use_cases/editor/section_editor.py
randy-seng: My bad @briemla ! I did not realize before making that comment that the if statement is supposed to check if the dictionary is empty. So, it was correct all along. Thus, I reverted your last comment applying this change request of mine.

randy-seng (approved):
Looks good! My last change request was wrong, thus I reverted your last commit 🙏🏻

PR #809, opened 2026-03-11
Add Playwright test for exporting road user assignments; implement CS…

StevenSchlechte:
OP#8353


briemla (changes requested):
See comment in OP

PR #821, opened 2026-03-27
bug/9548-incomplete-filenames-in-otanalytics-if-video-name-has-multiple-dots

randy-seng:
## Summary

Fixes save/export filename handling so generated paths and Save As dialogs consistently apply file extensions exactly once, including filenames with multiple dots and custom extensions such as `.otevents`.

## Changes

- Introduced `SavePathSuggestion` to centralize save directory, filename stem, context suffix, file type, and derived file path handling.
- Normalized extension handling so callers can pass file types with or without a leading dot.
- Fixed duplicated `.otevents` suffixes in CustomTkinter save dialogs by passing the filename without the final file type as `initialfile`.
- Fixed NiceGUI file extension parsing by stripping glob wildcards like `*` before building enforced suffixes.
- Refactored CustomTkinter export dialogs to return typed DTOs instead of exposing internal input dictionaries.
- Tightened export-count dialog behavior
- Updated unit and acceptance tests around save path suggestions, export dialogs, and file extension handling.

## Testing

- Updated unit tests for `SavePathSuggestion` integration and extension handling.
- Added coverage for NiceGUI file extension conversion.
- Improved acceptance-test timeout behavior so save regressions fail clearly instead of hanging.

OP#9548


schubsen (changes requested):
The files with the suffixes 
- .tracks_metadata.json
- .tracks.csv
- videos_metadata.json
still get truncated

schubsen (approved):

PR #832, opened 2026-05-08
bug/9664-road-user-assignment-highlighting-is-not-correct-when-loading-ottrk-before-otflow

randy-seng:
### Summary

This pull request addresses two improvements:

1. **Flow Highlighting Fix for `.ottrk` and `.otflow`**:  
   - Resolves an issue where road user assignment highlighting was not applied correctly when `.ottrk` files were loaded before `.otflow` files.  
   - Refactors `AddAllFlows` to validate all flows at once before insertion, ensuring consistency and avoiding duplicates.  
   - Introduces helper functions for validation and updates batch processing to be atomic.  
   - Updates flow-loading logic and expands test coverage.  

2. **Exclude Shared AI Coding Rules from Version Control**:  
   - Updates `.claude/rules/shared` to use symlinks rather than committing shared code style markdown files individually across repositories.  
   - This ensures consistent propagation of updates to shared rules without duplication.  

OP#9664


OTAnalytics/application/use_cases/flow_repository.py
```diff
@@ -32,41 +32,41 @@ def __call__(self, flow: Flow) -> None:
...
-                return True
-        return False
 
-    def is_flow_name_valid(self, flow_name: str) -> bool:
-        if not flow_name:
-            return False
-        return all(
-            stored_flow.name != flow_name
-            for stored_flow in self._flow_repository.get_all()
+def check_flow_already_exists(given: Flow, existing_flows: list[Flow]) -> None:
```
dahelb: Can also be an `Iterable[Flow]`, right (consistency with other type hints below)?

OTAnalytics/application/use_cases/flow_repository.py
```diff
@@ -93,9 +93,24 @@ def get(self) -> list[Flow]:
...
 class AddAllFlows:
-    def __init__(self, add_flow: AddFlow) -> None:
-        self._add_flow = add_flow
+    def __init__(self, flow_repository: FlowRepository) -> None:
+        self._flow_repository = flow_repository
 
     def add(self, flows: Iterable[Flow]) -> None:
-        for flow in flows:
-            self._add_flow(flow)
+        flow_list = list(flows)
```
dahelb, line 100: Is the list wrap here necessary? Satisfying `Iterable` should be enough, shouldn't it?

dahelb (dismissed):
Minor typing things. Rest looks good to me!

OTAnalytics/application/use_cases/flow_repository.py
randy-seng: Unfortunately calling `add_all` with exhaustable iterators such as generators, which are also subtypes of Iterable, will fail. This is why it is required to wrap the flows around list. Instead I will add another unit tests that covers this case.

dahelb (approved):

PR #835, opened 2026-09-04
Task/10256 add s3 connection and data transfer mode configuration

randy-seng:
## Summary

  Adds an S3 client layer and the startup configuration that selects where OTAnalytics gets
  its tracks and videos from. Nothing user-visible changes: with no configuration the
  transfer mode is `local-filesystem` and every existing path behaves exactly as before.
  This is the plumbing that the time-range loading slice (OP#10283) switches on, split out
  so it can be reviewed on its own.

  S3 access is configured entirely through environment variables. There is no config file
  and no new CLI flag.

  ## Changes

  - `TransferMode` (`domain/transfer_mode.py`) — `local-filesystem` (default) and `s3`.
    OTCloud's `ftp` member is deliberately absent, so an unsupported mode is rejected by
    the enum itself.
  - `application/startup_config.py` — the `StartupConfigError` hierarchy and
    `validate_transfer_mode(mode, start_webui)`. No S3 type.
  - `OTAnalytics/plugin_s3/` — `S3Connection`, `S3Download`, `S3Upload`, `S3ListObjects`,
    `S3Config`, `S3Env`, `parse_s3_config()`, `transfer_mode_from_env()`.
  - `ApplicationStarter` reads the mode, validates, then reads the S3 variables if needed.
  - `CONTEXT.md`, defining Transfer Mode, Startup Configuration, Key Prefix, User Source
    and Load Window.
  - `aioboto3` as a runtime dependency; `types-aiobotocore-s3` in dev.

  Configuration surface, six names shared with OTCloud's `S3Env`:

  | Variable | Required in s3 mode | Default |
  |---|---|---|
  | `DATA_TRANSFER_MODE` | always | `local-filesystem` |
  | `S3_BUCKET`, `S3_ACCESS_KEY`, `S3_SECRET_KEY`, `S3_KEY_PREFIX`, `S3_USER_SOURCE` | yes | — |
  | `S3_ENDPOINT_URL`, `S3_REGION` | no | `None` |
  | `S3_MAX_LOAD_DURATION` | no | `10h` |
  | `S3_DOWNLOAD_CONCURRENCY` | no | `8` |

  Three decisions worth knowing about.
  
  **The S3 client layer is copied from OTCloud, not shared with it.** OTCloud depends on
  `OTAnalytics==0.7.3` and already imports OTAnalytics internals, so moving these ~150 lines
  down here and having OTCloud import them was available and was rejected: it would couple
  OTCloud to this project's release cadence, so an S3 fix would need an OTAnalytics release
  and a pin bump before OTCloud could use it. The two copies will drift. That is the
  accepted cost — please don't unify them without revisiting
  `docs/adr/0001-duplicate-s3-layer.md`.

  **Nothing S3-specific reaches the application or domain layers.**
  `scripts/check_clean_architecture.py` documents domain → application → adapter → plugin
  with inward-only imports. `application/startup_config.py` deals only in `TransferMode`;
  `S3Config` and everything else S3 live in `plugin_s3`. `ApplicationStarter` is the single
  place that knows about both.

  **The mode check runs before the S3 variables are read.** `transfer_mode_from_env()`
  cannot fail on credentials, so the front-end rule is evaluated first. With them the other
  way round, a CTk user in S3 mode with no credentials was told which credentials were
  missing — when the real problem is that CTk cannot use S3 at all.

  ## Notes for the reviewer

  `plugin_s3/config/parsing.py` and `application/startup_config.py` carry the actual
  decisions. The four client primitives are near-verbatim copies of OTCloud's and can be
  skimmed.

  `S3Upload` is not called by anything in this branch. It is copied with the rest of the
  layer for the later result-upload work and is unit tested so it isn't sitting there
  unverified. Reasonable to argue it should wait — say so and I'll drop it.

  `ArgparseCliParser.__init__` had a mutable default argument
  (`arg_parser: ArgumentParser = ArgumentParser(...)`), so constructing it twice in one
  process raised `argparse.ArgumentError: conflicting option string: --cli`. Unrelated to
  S3, and the flag that made me hit it is gone, but the bug is real and the fix is kept.

  ## Risk and rollout

  Every startup now reads `DATA_TRANSFER_MODE`. **A stray `DATA_TRANSFER_MODE` in the
  environment will stop OTAnalytics for a non-S3 user** — plausible on a machine that also
  runs OTCloud, whose `TransferMode` includes `ftp`. It surfaces as a logged error naming
  the supported values and a clean return, not a traceback. If that trade is wrong, the fix
  is to warn and fall back in `transfer_mode_from_env()`.

  ## Testing

  Written test-first, each piece confirmed failing before it was made to pass.

  Also exercised by hand through `ApplicationStarter`, since the check-ordering problem was
  only visible end to end. Five paths: default resolves to `local-filesystem` with no S3
  config; CTk with `DATA_TRANSFER_MODE=s3` reports the `--webui` message; `--webui` with the
  same environment reports all five missing variables; `--webui` with complete credentials
  is accepted; and `DATA_TRANSFER_MODE=ftp` is rejected naming the supported values. All
  four failures log and return rather than raising.

  Not covered: nothing talks to a real S3 endpoint. The client layer is tested against
  `AsyncMock`, so wire-level behaviour — real pagination, real credentials, endpoint quirks
  — is unverified until the MinIO integration test lands with OP#10283.

OP#10256


OTAnalytics/plugin_s3/config/env_vars.py
```diff
@@ -0,0 +1,82 @@
...
+ENV_S3_MAX_LOAD_DURATION = "S3_MAX_LOAD_DURATION"
+ENV_S3_DOWNLOAD_CONCURRENCY = "S3_DOWNLOAD_CONCURRENCY"
+
+
+@dataclass
+class S3Env:
+    """S3 settings read from the environment.
+
+    Values are read when the instance is constructed. Names for the first six
+    fields match OTCloud's `S3Env` so a deployment already configured for
```
dahelb: Unsure if it wouldn't be better to prefix the variable names with `OTA_` or something similar.
This would make the separation clearer (which variable for which project) - we might want to have different settings (if not now, maybe later?).

OTAnalytics/plugin_s3/config/env_vars.py
dahelb, line 1: This whole file seems to be a bit verbose. I don't see the advantage of having a dataclass created from the environment variables first, and then feeding this into a parsing function in a different module. Could we maybe just merge the two modules (env_vars.py and parsing.py) and have a single method read the variables and check for completeness?

OTAnalytics/plugin_s3/config/parsing.py
```diff
@@ -0,0 +1,127 @@
...
+        bucket=_required(env.bucket),
+        region=env.region,
+        key_prefix=_required(env.key_prefix),
+        user_source=_required(env.user_source),
+        max_load_duration=_parse_max_load_duration(env),
+        download_concurrency=_parse_download_concurrency(env),
+    )
+
+
+def _required(value: str | None) -> str:
```
dahelb, line 113: simpler maybe: direct key access of `os.environ['MY_ENV']`, catch `KeyError`.
Easier to ask for forgiveness than permission ;)

OTAnalytics/plugin_s3/config/s3.py
```diff
@@ -0,0 +1,41 @@
...
+DEFAULT_DOWNLOAD_CONCURRENCY = 8
+
+
+@dataclass(frozen=True)
+class S3Config:
+    """Configuration for S3-compatible object storage.
+
+    Mirrors the `s3` block of the startup configuration one for one. Copied from
+    OTCloud's `S3Config` and deliberately kept separate from it — see
+    `docs/adr/0001-duplicate-s3-layer.md`. OTCloud's `otcloud_source` is dropped;
```
dahelb: doc isn't commited, so this reference should be removed imo

OTAnalytics/plugin_s3/config/s3.py
```diff
@@ -0,0 +1,41 @@
...
+            live. Fixed at startup; the user selects a time range, not a prefix.
+        user_source (str): base path on the local filesystem that downloaded
+            objects are written to, mirroring their S3 keys. Named after
+            OTCloud's `S3_USER_SOURCE` for cross-repo consistency.
+        max_load_duration (timedelta): longest time range that may be loaded in
+            one go. A longer selection is clamped to it.
+        download_concurrency (int): how many objects to download at once.
+    """
+
+    endpoint_url: str | None
```
dahelb: Unsure if we should not rather make this required.
I am leaning torwards requiring the endpoint_url. I think the danger of misconfiguration due to a missing (or misnamed) parameter and then falling back more or less silently to AWS is bigger then the minor inconvenience of having to specify the AWS endpoint specifically if a user should really want this.

OTAnalytics/plugin_s3/download.py
dahelb, line 1: Can we merge the download, upload, list functionality under a common class? having three different ones with identical constructors seems a bit redundant.
Any advantage in keeping them separate?

dahelb (commented):
Some remarks, mostly regarding code style, some of which are debatable. Nothing really blocking a merge, but I would be interested in what you think.

OTAnalytics/plugin_s3/config/env_vars.py
randy-seng: Thanks for the feedback. I knowingly accepted the verbosity for better code readability and testability.

Reading the env vars and the validating/parsing the configurations are two distinct operations (separation of concerns). The dataclass makes it clear what data is available from the environment, whereas parsing focuses only on validating and building the final config.

Having them separated promotes clarity, since the user can see exactly what env vars are read and what gets validated. A single merged function would hide this. 

This also makes testing easier, since we don't need to mock os.environ and can just dependency inject the needed object.

OTAnalytics/plugin_s3/config/parsing.py
randy-seng: I now get the connection with the other comment in env_vars.py. This change is only possible if we apply your feedback and get rid of the S3Env dataclass entirely and use `os.environ` directly.

This function's sole purpose is only to conform to the type checker, since the validation for the existence of all required env vars already happened in `parse_s3_config`.

OTAnalytics/plugin_s3/config/env_vars.py
randy-seng: Agreed! Done!

OTAnalytics/plugin_s3/config/s3.py
randy-seng: Done :)

OTAnalytics/plugin_s3/config/s3.py
randy-seng: Done!

OTAnalytics/plugin_s3/download.py
randy-seng: Yes, I agree with the identical constructors being redundant, there is a way to simplify it further which I now did. But would still keep the use cases separate to enforce the integration segregation principle, which basically states that no client code should be forced to depend on methods or properties it does not use.  For example we don't want a use case that only needs read privileges (download, list) to also have write privileges (upload) if it doesn't need to.  Keeping them separate also lets each other evolve independently without touching the others.

dahelb (dismissed):
Thanks for your comments!!

dahelb (approved):
re-approve

PR #836, opened 2026-09-04
OP#10280: Make the source of tracks and videos interchangeable

randy-seng:
## Summary

  `DummyViewModel.load_tracks` and `add_video` had their file-chooser calls inline, so where
  input files come from was fixed at the view model. This introduces `ProvideTrackFiles` and
  `ProvideVideoFiles` with local-filesystem implementations and moves that dialog code into
  them. Behaviour is unchanged — this is preparation for loading tracks and videos from S3
  (OP#10283), split out so the refactor can be reviewed without the feature on top of it.

  `DummyViewModel` is net −29 lines and no longer contains any file-chooser knowledge.

  ## Changes

  - `application/use_cases/provide_input_files.py` — `ProvideTrackFiles` and
    `ProvideVideoFiles`, each `async def provide(self) -> list[Path]`.
  - `adapter_ui/local_file_providers.py` — `LocalTrackFileProvider` and
    `LocalVideoFileProvider`, holding the dialog code lifted out of the view model.
  - `gui_application.py` — two new `cached_property` members wiring them in.
  - `SUPPORTED_VIDEO_FILE_TYPES` moved from `dummy_viewmodel.py` to `application/config.py`.

  Two decisions worth knowing about.

  **Each provider owns its own dialog interaction, rather than being handed a selection.**
  That is what keeps the view model mode-agnostic: `load_tracks` is now
  `files = await provider.provide()` followed by `application.add_tracks_of_files(files)`,
  and it never learns that anything other than a local file chooser exists. The alternative
  — the view model choosing a dialog and passing the result to a provider — would have put a
  branch on transfer mode into `DummyViewModel`.

  **`SUPPORTED_VIDEO_FILE_TYPES` went to `application/config.py`, not with the code that
  moved.** It states what the application can read, not how files get chosen.
  `tests/utils/playwright_helpers.py` already used it as a plain fact about videos, and the
  S3 video provider will need the same list without importing from an adapter it has nothing
  to do with.

  ## Notes for the reviewer
  
  Start with `adapter_ui/local_file_providers.py` and the two view model methods; everything
  else is wiring.

  **`UiFactory` and `ViewModel` are deliberately untouched** — both files have empty diffs.
  OTCloud's `plugin_plotter/track_plotter.py:48` defines `UnimplementedUiFactory(UiFactory)`
  implementing every abstract method, so adding one would make that class abstract and raise
  `TypeError` on instantiation over there. Worth preserving in review.

  The strongest evidence this is behaviour-preserving is that **no existing test needed
  changing**. The one pre-existing file touched is `tests/utils/playwright_helpers.py`, and
  only its import line, which followed the constant to its new home.

  Two deliberate deletions rather than moves. `load_tracks` used to do
  `track_paths = [Path(file) for file in track_files]`; `UiFactory.askopenfilenames` is
  declared `-> list[Path]` and both implementations honour it, so that was `Path(Path(...))`.
  And the ~15 lines building `extension_options` are reproduced in
  `LocalVideoFileProvider._video_extension_options()` with the same insertion order —
  `"All File Endings"` first, then one entry per extension.

  `test_dummy_viewmodel_loading.py` stubs `_start_action`/`_finish_action` on the instance.
  `load_tracks` carries `@action`, whose `_finish_action` refreshes the canvas and re-enables
  buttons — a fully wired widget tree, for reasons unrelated to file loading. The stubs are
  documented in the test. They do mean a regression that throws *after* delegation would go
  unnoticed by these tests specifically; the playwright suite covers the action lifecycle.

  ## Risk and rollout

  `DummyViewModel.__init__` gains two required parameters. Nothing outside this repo
  constructs it — checked OTCloud, excluding `.venv` and `.claude/worktrees`. OTCloud's
  `PlotApplicationStarter` and `LiveApplicationStarter` both subclass
  `OtAnalyticsGuiApplicationStarter` without overriding `view_model`, so they inherit the new
  `cached_property` members automatically. Both override `ui_factory` with
  `UnimplementedUiFactory`, which is harmless: the providers only store the reference and
  never call it headless.

  No config, no migration. Reverting is one commit.

  ## Testing

  - `uv run pytest tests/unit` — 1286 passed, 6 skipped (1276 on `main`, +10 new)
  - `uv run pytest tests/system` — 2 passed
  - `uv run pytest tests/regression` — 6 passed, 4 skipped
  - `uv run mypy OTAnalytics tests` — clean, 535 files
  - `uv run python scripts/check_clean_architecture.py` — 3 violations, the same 3 as on
    `main`
  - All pre-commit hooks passed

  Written test-first, each piece confirmed failing before it was made to pass.

  Still to verify by hand: nobody has clicked "Add tracks…" or "Add videos…" in either
  front-end since the change. The unit tests pin that the providers pass the same dialog
  arguments the view model used to, and the acceptance suite exercises the flow, but a real
  click in the CTk GUI and the webui would be worth doing before merge.

  OP#10280


OTAnalytics/adapter_ui/local_file_providers.py
```diff
@@ -0,0 +1,62 @@
+"""Local-filesystem implementations of the input file providers.
+
+These hold the file-chooser interaction that used to sit inline in
+`DummyViewModel`, so that swapping in a different source of input files does not
+touch the view model at all.
```
schubsen: Please remove the comment about what changed in the file description.

OTAnalytics/application/use_cases/provide_input_files.py
```diff
@@ -0,0 +1,36 @@
+"""Obtaining the track and video files a user wants loaded.
+
+The seam that lets input files come from somewhere other than the local
+filesystem. Each implementation owns its own way of asking the user what to
+load, so nothing downstream — neither `LoadTrackFiles` nor the view model —
+needs to know where the files came from.
```
schubsen: Should the description reference class names like `LoadTrackFiles` or the view model if the class does not know them? In my opinion this should be removed.

schubsen (changes requested):

schubsen (dismissed):

schubsen (approved):

PR #844, opened 2026-09-12
OP#10325: Start the web UI when a configuration file is given

randy-seng:
## What this does

`OTAnalytics --webui --config <file>` never started. The process died before binding a port, with a traceback and no application.

`VisualizationFiltersForm` registers itself with the view model in `__init__`, but its widgets are created in `build()`, which runs only when a browser asks for the page. The `--config` preload runs between those two points — inside the `webserver` property, before any page is built — so it notified a form whose widgets did not exist.

## Notes for the reviewer

The fix is wider than the one crash. **Ten** methods on that form reach a widget created only in `build()`, and each would fail the same way; `update_date_range` was just the one on the preload path. The tests call all ten. That is why "not yet built" became a state the form has — declared `None` in `__init__`, the way the navigation buttons already were — rather than a `hasattr` guard on the single method that happened to crash.

Same family as the bug found while reviewing OP#10321, where `info_box` reached for a browser client in this same startup window.

## Limitation

A **malformed** otconfig still stops startup. `ApplicationStarter.run_config` (`application_starter.py:72`) parses the file to learn the log file and save paths, before any of this exists, and that parse is unguarded. Only failures in the second, preloading load are caught: unreadable tracks, clashing sections, a reference resolving nowhere.
Making an unparseable file survivable means letting `RunConfiguration` be built without one — bigger than this defect, and worth its own ticket.

## Testing

By hand: `python -m OTAnalytics --webui --config project.otconfig` now starts and serves.

Also verified against the blocked case that prompted this — merging this branch into `task/10322` on a throwaway branch makes `tests/acceptance/test_load_from_s3.py` pass against MinIO, 2 of 2. Expect one merge conflict there: slice 6's user-source wipe sits on the lines next to the new `preload_project()` call.

OP#10325

### DIFF
--- a/OTAnalytics/plugin_ui/nicegui_application.py
+++ b/OTAnalytics/plugin_ui/nicegui_application.py
@@ -2,6 +2,7 @@ from functools import cached_property
 from typing import Protocol
 
 from OTAnalytics.adapter_ui.ui_factory import UiFactory
+from OTAnalytics.application.logger import logger
 from OTAnalytics.domain.progress import ProgressbarBuilder
 from OTAnalytics.plugin_progress.tqdm_progressbar import TqdmBuilder
 from OTAnalytics.plugin_prototypes.track_visualization.track_viz import (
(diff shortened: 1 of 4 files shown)


OTAnalytics/plugin_ui/nicegui_application.py
```diff
@@ -103,14 +104,32 @@ def webserver(self) -> Webserver:
...
-        self.preload_input_files.load(self.run_config)
+        self.preload_project()
         return NiceguiWebserver(
             page_builders=[main_page_builder],
             layout_components=NiceguiLayoutComponents(),
             hostname=DEFAULT_HOSTNAME,
             port=DEFAULT_PORT,
         )
 
+    def preload_project(self) -> None:
```
randy-seng, line 114: @schubsen Should we fail hard open preloading a faulty otconfig or should we nonetheless spin up the OTAnalytics app?

schubsen: Fail hard. It is one command, e.g. `uv run -m OTAnalytics --webui --config <otconfig file>`, that leads to this behavior. It should either run or fail without unexpected side effects (WebUI started).

tests/unit/OTAnalytics/plugin_ui/nicegui_gui/pages/visualization_filters_form/test_container.py
```diff
@@ -0,0 +1,92 @@
+"""Tests the form before its page has been built.
```
randy-seng, line 1: Additional information: 

The form introduces itself to the view model in `__init__`
(`_introduce_to_viewmodel`), so the view model can call it from the moment it
exists. `build()` runs much later, when a browser asks for the page. Anything
the application does in between -- preloading a `--config` file while the
webserver is still being constructed, for one -- reaches a form whose widgets
do not exist yet.

schubsen (approved):

